PR #8, opened 2022-11-16
fix(refs T30074): Resolve Path Issues

salisdemos:
Just clean up all the path declarations. Something got wrong during the move.

moved prefixClass and prefixClassMixin from 'demosplan-ui` to `demosplan-utils` to avoid a cross referencing (in demosplan-ui its now marked as deprecated, but not removed)

**Ticket** https://yaits.demos-deutschland.de/T30074

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Change Log
+
+
+## v0.0.2
+2022-11-16
+
+### Bug Fixes
+- Fix path declarations to have a buildable package
+
+## v0.0.1 
+2022-11-15
+
+- This is unfinished and broken. We just need a startingpoint
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Bug Fixes
 - Fix path declarations to have a buildable package
 
+### Features
+- Deprecate prefixClass and prefixClassMixin
+
 ## v0.0.1 
 2022-11-15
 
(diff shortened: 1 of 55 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Change Log
+
+
+## v0.0.2
```
graupnerdemos, line 4: Technically, when writing a changelog, the top most version should be something like **unreleased** and adjusted to the desired version on publish. This avoids conflicts when changes come from multiple PRs.

package.json
```diff
@@ -1,11 +1,13 @@
 {
   "name": "@demos-europe/demosplan-ui",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "license": "MIT",
   "private": false,
   "description": "Vue components, Vue directives, Design Token and Scss files to build interfaces for demosPlan.",
   "dependencies": {
     "@braintree/sanitize-url": "^6.0.1",
+    "@demos-europe/demosplan-ui": "^0.0.2",
```
graupnerdemos, line 9: Why does this package have itself as a dependency now?

graupnerdemos (changes requested):

CHANGELOG.md
salisdemos: Thanks for the Advise. I would say we should add that next Time.

package.json
salisdemos: Because even before the package was required. But I just didn't checked why, but only recognized it. Stupid thing though. Changed it in 0a132519c3a9037d28a1a42f88e5c57f3896d2f1

graupnerdemos (approved):

PR #10, opened 2022-11-18
add npm entrypoint and remove other index.js files

muellerdemos:
Remove multiple index.js entryfile in favor of having just one on root level

- Since it appears to be problematic to have multiple entrypoints, this was changed to a single index.js
- This means all demosplan-ui components can be imported directly via the package, without specifying the exact location of components or javascript files within the npm package


spiess-demos (changes requested):
i'd propose to keep the index files at the certain directories, but then export it all with the root index file.

muellerdemos:
Sorry for the confusion @spiess-demos this PR is actually obsolete by now. Since the process changed a little

PR #11, opened 2022-11-22
Remove license and doc blocks

spiess-demos:
As there is no requirement for license file headers within demosplan-ui, these can be removed. Also, to prepare components for documentation, existing docs are moved to `*.stories.mdx` files.

Best reviewed commit wise, as the removal of the license headers happens within separate commits.


ahmad-demos (approved):
lgtm, thanks!

sakutademos (approved):
thanks a lot! lgtm!

mussbach:
Why is there no requirement for the headers?

spiess-demos:
> Why is there no requirement for the headers?

As i understood @graupnerdemos - and to my knownledge - the MIT license requires that...

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software. 

As we include a LICENSE file with the repo, this requirement is fulfilled. At least, most other repos that are MIT licensed interpret it that way).

PR #12, opened 2022-11-22
fix_add_webpack_config_to_have_a_buildstep_for_demosplan_ui

muellerdemos:
This webpack config enables us, to have a proper build step for npm packagin the demosplan-ui components.

PR #13, opened 2022-11-22
f_create_demosplan_component_library

muellerdemos:
This PR enables in combination with this one https://github.com/demos-europe/demosplan-ui/pull/12
the creation of a vue component library that can be installed via npm. 

To achieve this, mostly some paths had to be adjusted. 

The biggest issue in this PR was the initial version of vue-loader which was simply not compatible with vue 2.7.  Hence a downgrade to v15 was essential. Before the downgrade it was simply not possible to run the build process which also threw some strange errors along the way like styles were undefined, which didn't make any sense.

However now it is possible to create an umd.js containing all demosplan-ui components, which should be usable directly out of the npm package.

To create all components at once and also to be able to use all components directly out of the package, all files in` /core` and `/components`  were added to the components index.js file

To test this, simply move into the demosplan-ui folder and run 'yarn build'

Please note that there are still some minor warnings occuring, which were cause currently by the demosplan-js-utils. However I also found a solution for this and will add this following PR as soon as its created here:


README.md
```diff
@@ -0,0 +1,15 @@
...
+npm run build
+```
+
+or
+
+```
+yarn build
+```
+
+to bundle your application
```
salisdemos, line 15: we should add some - at least basic - meaningfull informations here.

README.md
```diff
@@ -0,0 +1,15 @@
+# 🚀 Welcome to your new awesome project!
+
+This project has been created using **webpack-cli**, you can now run
```
salisdemos, line 3:
```suggestion
# DEMOSPLAN UI

Vue components, Vue directives, Design Token and Scss files to build interfaces for demosPlan.
```

components/DpLabel/DpLabel.vue
```diff
@@ -23,9 +23,9 @@
 </template>
 
 <script>
-import { CleanHtml } from '../../directives'
+import { CleanHtml } from '../../directives/CleanHtml/CleanHtml'
```
salisdemos, line 26: do we want to import like this here or like `import { CleanHtml } from '../../'` ?
I see both versions in this PR

components/core/DpDataTable/DpTableHeader.vue
```diff
@@ -1,5 +1,14 @@
+<license>
+  (c) 2010-present DEMOS E-Partizipation GmbH.
+
+  This file is part of the package @demos-europe/demosplan-ui,
+  for more information see the license file.
+
+  All rights reserved
+</license>
+
```
salisdemos, line 9: Didn't we want to drop the license-header?

dist/main.js
```diff
@@ -0,0 +1,2 @@
+/*! For license information please see main.js.LICENSE.txt */
```
salisdemos, line 1: shouldn't this be excluded from git? I have seen its configured like so. but then it shouldnt be here too

package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "@demos-europe/demosplan-ui",
-  "version": "0.0.2",
+  "name": "my-webpack-project",
+  "version": "1.0.0",
   "license": "MIT",
   "private": false,
-  "description": "Vue components, Vue directives, Design Token and Scss files to build interfaces for demosPlan.",
+  "description": "My webpack project",
```
salisdemos, line 6: "description": "Vue components, Vue directives, Design Token and Scss files to build interfaces for demosPlan.",

package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "@demos-europe/demosplan-ui",
-  "version": "0.0.2",
+  "name": "my-webpack-project",
```
salisdemos, line 2: "name": "@demos-europe/demosplan-ui",

package.json
```diff
@@ -1,12 +1,12 @@
...
+  "version": "1.0.0",
```
salisdemos, line 3: "version": "0.0.2",

package.json
salisdemos, line 6: I am pretty sure that solething like this is missing to make it usable in projects:
```
  "main": "./dist/demosplan-ui.umd.js",
  "module": "./dist/demosplan-ui.es.js",
  "exports": {
    ".": {
      "import": "./dist/demosplan-ui.es.js",
      "require": "./dist/demosplan-ui.umd.js"
    }
  },
  ```
  (at least "main")

**Review by salisdemos — CHANGES_REQUESTED** [review]
Thanks a lot for the great effort. 
I still can't build the package, but I think we are close. 

- We have to get rid of the import from the store (components/core/DpEditor/DpBoilerPlateModal.vue:35)
- `demosplan-utils` misses `fscreen` and `tooltip.js` as dependencies.

File: components/DpLabel/DpLabel.vue (hunk 3 above)
Comment (spiess-demos) [reply components/DpLabel/DpLabel.vue:26]: i like the '../../' more. even more i'd like some `@components` / `@directives` alias but thats a thing for another day.

**Comment by spiess-demos** [comment]
> * `demosplan-utils` misses `fscreen` and `tooltip.js` as dependencies.

tackled in https://github.com/demos-europe/demosplan-js-utils/pull/5

File: components/core/DpDataTable/DpTableHeader.vue (hunk 4 above)
Comment (mussbach) [reply components/core/DpDataTable/DpTableHeader.vue:9]: why should we?

Comment (graupnerdemos) [reply components/core/DpDataTable/DpTableHeader.vue:9]: It's not necessary to have it in there for MIT compliance and it has been known to cause issues with build tooling. That's why we decided to drop it.

Comment (graupnerdemos) [reply components/core/DpDataTable/DpTableHeader.vue:9]: But this is demosplan and not MIT, thus the header has to stay.

Comment (mussbach) [reply components/core/DpDataTable/DpTableHeader.vue:9]: Ok. This needs to be documented somewhere prominently that we assessed this and dropped it. Is this already the case?

Comment (muellerdemos) [reply components/core/DpDataTable/DpTableHeader.vue]: removed this leftover license tag in [9754f51](https://github.com/demos-europe/demosplan-ui/pull/13/commits/9754f51810b3dfa98093e5d8e5f7243f4d11ed87)

File: dist/main.js (hunk 5 above)
Comment (muellerdemos) [reply dist/main.js]: yes I added it to gitignore but after the file was already pushed. this was removed here [d2c5c7e](https://github.com/demos-europe/demosplan-ui/pull/13/commits/d2c5c7eb50a67e695834cd1a9f3955378baf726c)

File: package.json (hunk 6 above)
Comment (muellerdemos) [reply package.json]: this and the following concerns in relation to package.json have been adjusted here [e72971f](https://github.com/demos-europe/demosplan-ui/pull/13/commits/e72971fc72bf1d65b82123a9ad58a4ab23b167e2)

File: README.md (hunk 2 above)
Comment (muellerdemos) [reply README.md]: thanks, added a readme here [f04d3e4](https://github.com/demos-europe/demosplan-ui/pull/13/commits/f04d3e42aaac35bedbd18112bef7d629108083ee)

File: webpack.config.js [hunk 9]
```diff
@@ -12,7 +12,7 @@ function resolve (dir) {
 
 const config = {
   entry: [
-    resolve('./components/index.js')
+    resolve('./index.js')
```
spiess-demos, line 15: i think there is no need to have the index file here. instead, i'd also move it into /src, and specify that here like 

```suggestion
    resolve('./src/index.js')
```

this way you can also drop the `./src/` prefixes in every import statement in the index.js.

index.js
```diff
@@ -1,32 +1,155 @@
...
-export {
-  CleanHtml,
+import { exactlengthHint, maxlengthHint, minlengthHint } from './src/utils/'
+
+import {DpButton} from './src/components/'
+import {DpDetails} from './src/components/'
+import {DpIcon} from './src/components/'
+import {DpInput} from './src/components/'
+import {DpLabel} from './src/components/'
+import {DpLoading} from './src/components/'
```
spiess-demos, line 12:
```suggestion
import {
    DpButton,
    DpDetails,
    DpIcon,
    DpInput,
    DpLabel,
    DpLoading
} from './src/components/'
```

spiess-demos (changes requested):
the paths in [.storybook/main.js:3:5 ](https://github.com/demos-europe/demosplan-ui/blob/main/.storybook/main.js#L3-L5) need to be adapted as well.

webpack.config.js
muellerdemos: I think so, too. Done her [88792b4](https://github.com/demos-europe/demosplan-ui/pull/13/commits/88792b4c918516a7b14bacc779208738a24ed628)

index.js
muellerdemos: [64abf45](https://github.com/demos-europe/demosplan-ui/pull/13/commits/64abf452a7b1eca007906c680653a8916dbd8bc0) however not fully sorted alphabetically yet

PR #17, opened 2022-11-25
fix: use correct hook for now

spiess-demos:
until we publish via npm, the prepublish hook does not fire. this can be rolled back later.


salisdemos (approved):
thanks. Should do the trick

PR #18, opened 2022-11-25
fix: Change Hook to build dist

salisdemos:
Since "postinstall" looks like the wrong hook. Lets try "postcompress"


muellerdemos (approved):
lgtm at this point

PR #19, opened 2022-11-25
fix: Make posthook Work without publishing the Package

salisdemos:
though we need a working Dev, we can't wait to publish all the time. Therefore we need something that works on install.


graupnerdemos (approved):
lgtm

PR #21, opened 2022-11-29
fix: Make async Components work again.

salisdemos:
We had to make tiptap a peerDependency, so prosemirror ca be resolved propperly in dplan-core 

Declaring the Directives within shouldn't part of this PR, but When I try to make a new one, without these Changes , it looks like it doesn't work anymore.

You find the Key-Changes in the "webpack.config".

Related in dplan-core: https://github.com/demos-europe/demosplan-core/pull/387


spiess-demos (approved):
code review only, however this looks legit and i think it also enables tree shaking for demosplan-ui.

src/components/core/DpEditor/DpEditor.vue
```diff
@@ -357,14 +357,12 @@ import {
   TableRow,
   Underline
 } from 'tiptap-extensions'
-
+import { CleanHtml, Tooltip } from '../../../directives/'
```
muellerdemos, line 360:
```suggestion
import { CleanHtml, Tooltip } from '../../../directives'
```
It kind of feels odd to nitpick this, since I was the one adding those slashes in the first place, however, the slash should be removed, sorry.

src/components/core/DpEditor/DpRecommendationModal/DpInsertableRecommendation.vue
```diff
@@ -40,7 +40,7 @@
 </template>
 
 <script>
-import { CleanHtml } from '../../../../directives/'
+import { CleanHtml, Tooltip } from '../../../../directives/'
```
muellerdemos, line 43: same nitpick as above

muellerdemos (changes requested):
Sorry for nitpicking this

src/components/core/DpButtonIcon.vue
```diff
@@ -14,6 +14,8 @@
 </template>
 
 <script>
+import { Tooltip } from './../../directives'
```
hwiem, line 17: Here and below: import can be shortened

```suggestion
import { Tooltip } from '../../directives'
```

src/components/core/DpEditor/DpBoilerPlate.vue
```diff
@@ -45,6 +45,7 @@
 <script>
 import { CleanHtml } from '../../../directives/CleanHtml/CleanHtml'
 import DpMultiselect from '../form/DpMultiselect'
+import { Tooltip } from '../../../directives'
```
hwiem, line 48: We should use one import statement for CleanHtml and Tooltip:
`import { CleanHtml, Tooltip } from '../../../directives'`

src/components/core/DpEditor/DpEditor.vue
hwiem, line 360: Don't we have to add 
`tootip: Tooltip`
to `directives`?

src/components/core/DpTooltipIcon.vue
```diff
@@ -6,6 +6,8 @@
 </template>
 
 <script>
+import Tooltip from './../../directives/Tooltip/Tooltip'
```
hwiem, line 9:
```suggestion
import { Tooltip } from '../../directives'
```

And don't we have to add
`tooltip: Tooltip`
to `directives`?

src/components/core/DpTreeList/DpTreeListNode.vue
```diff
@@ -103,6 +103,7 @@ import DpIcon from '../../DpIcon/DpIcon'
 import DpTreeListCheckbox from './DpTreeListCheckbox'
 import DpTreeListToggle from './DpTreeListToggle'
 import draggable from 'vuedraggable'
+import { Tooltip } from './../../../directives'
```
hwiem, line 106:
```suggestion
import { Tooltip } from '../../../directives'
```

hwiem (changes requested):

hwiem (approved):
lgtm!

src/components/DpLabel/DpLabel.vue
```diff
@@ -23,15 +23,16 @@
...
 import { de } from '../shared/translations'
 import { prefixClassMixin } from '@demos-europe/demosplan-utils'
 
 export default {
   name: 'DpLabel',
 
   directives: {
-    cleanhtml: CleanHtml
+    cleanhtml: CleanHtml,
+    tootip: Tooltip
```
hwiem, line 35: The approve was premature, there's a typo :see_no_evil: (here and other occurrences)
```suggestion
    tooltip: Tooltip
```

salisdemos: luckily its still a draft ;-)

salisdemos: fixed in 54060a05b7048f10d09efe3170dcb3ed76f0911a

package.json
```diff
@@ -52,10 +50,16 @@
...
-    "vuex": "^3.6.2",
     "webpack": "^5.75.0",
     "webpack-cli": "^5.0.0"
   },
+  "peerDependencies": {
+    "tiptap": "^1.32.2",
+    "tiptap-commands": "^1.17.0",
+    "tiptap-extensions": "^1.35.2",
+    "vue": "^2.7.14",
+    "react": "16.14.0"
```
spiess-demos, line 61: why do we depend on react here?

salisdemos: because of storybook.

spiess-demos: ok but storybook comes with its own package.json where it defines its dependencies... and as demosplan-ui does not require react directly, i do not understand why we have to define it as a peer dependency here.

salisdemos: maybe we don't need to. But after adding it, all the warnings disappeared

muellerdemos (approved):
lgtm

PR #33, opened 2022-12-14
fix: Make async Components work again.

salisdemos:
We had to make `tiptap` a peerDependency, so `prosemirror` ca be resolved properly in `demosplan-core`

Declaring the Directives within shouldn't part of this PR, but When I try to make a new one, without these Changes , it looks like it doesn't work anymore.

Related in dplan-core: https://github.com/demos-europe/demosplan-core/pull/387


salisdemos:
nice try.

PR #34, opened 2022-12-16
Allow target blank in v-cleanhtml

spiess-demos:
As we allow the target attr, we have to make sure that the link is safe. Chromium seems to add `rel="noopener"` automatically as of v88, however to also ensure correct behavior for other engines, the `rel` attr is added whenever a target is set to "_blank".

See
- https://github.com/cure53/DOMPurify/issues/317
- https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/

### DIFF
--- a/src/directives/CleanHtml/CleanHtml.js
+++ b/src/directives/CleanHtml/CleanHtml.js
@@ -9,7 +9,8 @@ import DOMPurify from 'dompurify'
  * @see https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/
  */
 DOMPurify.addHook('afterSanitizeAttributes', function (node) {
-  if (!'target' in node) {
+  const nodeHasTarget = 'target' in node
+  if (!nodeHasTarget) {
     return
   }
   const target = node.getAttribute('target')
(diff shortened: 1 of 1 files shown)


salisdemos (approved):
looks good. Thank You

src/directives/CleanHtml/CleanHtml.js
```diff
@@ -1,5 +1,24 @@
...
+ * As we allow the target attr, we have to make sure that the link is safe.
+ * Chromium seems to add `rel="noopener"` automatically as of v88, however
+ * to also ensure correct behavior for other engines, the `rel` attr is added
+ * whenever a target is set to "_blank".
+ *
+ * @see https://developer.chrome.com/docs/lighthouse/best-practices/external-anchors-use-rel-noopener/
+ */
+DOMPurify.addHook('afterSanitizeAttributes', function (node) {
+  const nodeHasTarget = 'target' in node
+  if (!nodeHasTarget) {
```
salisdemos, line 13: why not 
```
if ('target' in node === false) {
```
?

spiess-demos: this way, ui suggests to "simplify the expression" which i did with b665fb76e4703c634fccd754d9b17dacaeacb640

PR #36, opened 2022-12-16
Configure storybook to use Webpack 5

spiess-demos:
To mitigate issues arising from Webpack 4 (Storybook builder default) being installed alongside Webpack 5 (being the build tool for demosplan-ui), Storybook is configured to use Webpack 5, too.

Along the way, Storybook is upgraded to its latest version.

See:
- https://github.com/vuejs/vue-cli/issues/5986
- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#fixing-hoisting-issues


muellerdemos (approved):
lgtm - tested + code review

PR #48, opened 2022-12-22
refactor(refs T29004): pass label props as object

hwiem:
As was implemented for other form components previously, the label props are now passed in a single object. That way it's easier to understand what they are used for.

**Breaking**: The `label` prop expects an `object` now (previously a `string`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## UNRELEASED
 
+- **Breaking:** Refactor DpRadio props: bundle label related props in a single object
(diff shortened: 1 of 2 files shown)


spiess-demos (approved):
I ask myself if today is the day we move the `label` prop into `src/shared/props/props.js`. Apart from that, lgtm!

hwiem:
> I ask myself if today is the day we move the `label` prop into `src/shared/props/props.js`. Apart from that, lgtm!

I'd like to keep this PR small, but moving it there can be a next step

spiess-demos (approved):

salisdemos (approved):

PR #49, opened 2022-12-22
Optimize bundling by using webpack externals

spiess-demos:
At the moment, demosplan-ui is bundled into one huge file (with some async files split out). Many 3rd party modules are used by demosplan-ui as well as demosplan-core, leading to that code being bundled twice within demosplan-core.

While it is not the final solution (which would involve a tree-shakeable demosplan-ui), [webpack allows to manually define externals](https://webpack.js.org/configuration/externals/) which are then not bundled within the build.

This PR excludes all modules shared between demosplan-ui and demosplan-core. Tiptap is not excluded, as demosplan-core does not depend on it anymore.

Testing this locally is a little bit complicated due to the current build/publishing process. You would have to...

- Check out this branch and duplicate it with another name
- Change `package.json` to have webpack, webpack-cli, and style-dictionary as dependencies instead of devDependencies
- Change `package.json` scripts to use "postinstall" instead of "prepublish" as build hook
- Push it & copy last commit hash

Then, within demosplan-core and inside container:

- change package.json to use `"@demos-europe/demosplan-ui": "github:demos-europe/demosplan-ui#<last-commit-hash>",`
- `rm -rf node_modules`
- `yarn cache clean`
- `yarn install`
- `yarn prod:<project> --analyze`

Then, you should be able to click around and see if everything works. BundleAnalyzer Data shall be generated within `client/stats/bundle_analysis_<projectname>_development-html`.


muellerdemos (approved):
reviewed in pair session, this lgtm and we could verify the functionality of this PR

Thanks a lot for spotting our big chunks and for providing this solution!

PR #53, opened 2023-01-02
feat(refs T21119): Add a wrapper for draggable

gruenbergerdemos:
Add a wrapper for draggable, so it is easier to maintain.

Related PR: https://github.com/demos-europe/demosplan-core/pull/68

For testing see https://github.com/demos-europe/demosplan-ui/pull/49


src/components/core/DpDataTable/DpDataTable.vue
```diff
@@ -1,19 +1,19 @@
...
 import DpLoading from '../../DpLoading/DpLoading'
 import DpTableHeader from './DpTableHeader'
 import DpTableRow from './DpTableRow'
-import draggable from 'vuedraggable'
 
 export default {
   name: 'DpDataTable',
 
   components: {
+    DpDraggable,
```
spiess-demos, line 13: If async imported, only DataTable instances that are draggable, load the additional lib. 

```
DpDraggable: () => import('../DpDraggable')
```

In the `DpTreeList` and `DpTreeListNode` components it is a little more complicated, as those use the `<dp-draggable>` element also if not draggable. We would have to work with `<component :is="" />` within those templates to be able to actually split out the draggable lib from the main bundle.

gruenbergerdemos: I don't quite understand this. What should I put inside `<component :is="condition" />`? On what should it be dependent?

spiess-demos: If DpTreeList/DpTreeListNode is not draggable (via its `draggable` prop), it would not have to load the DpDraggable component. So we could set the component that renders the list like so: 

```
<component :is="draggable ? 'dp-draggable' : 'div' />
```

This would allow us to `import()` DpDraggable, reducing the size of the bundled demosplan-ui.

gruenbergerdemos: Ah, now I understand, thanks. Done in [6b82ae8](https://github.com/demos-europe/demosplan-ui/pull/53/commits/6b82ae815e4fb8677bf54bcc3f461efbe2cd771b)

spiess-demos: Thank you for approaching this! I checked out the branch to also add the lazy imports `DpDraggable: () => import('../DpDraggable'),` in DpTreeList and DoTreeListNode, however this does not lead to DpDraggable being split out when building all of it. I don't know whats the problem here, so i assume further investigation is needed. However to not further block the merge of this PR, i think this may as well be tackled later.

src/components/core/DpTreeList/DpTreeListNode.vue
```diff
@@ -53,17 +53,18 @@
         v-else
         class="min-width-25" />
     </div>
-    <draggable
-      v-model="tree"
-      v-bind="options.draggable"
+    <dp-draggable
+      :drag-across-branches="options.dragAcrossBranches"
       class="list-style-none u-mb-0 u-1-of-1"
-      :disabled="hasDraggableChildren === false"
```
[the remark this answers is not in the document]
gruenbergerdemos: I couldn't find any case in which it would be necessary, but since it also works and I may have forgotten a case, I changed it in [fb1eab7](https://github.com/demos-europe/demosplan-ui/pull/53/commits/fb1eab79f4aa682b238c25be74f67160e10c5542).

salisdemos (approved):
I think after resolving the conflicts. Its fine now

PR #55, opened 2023-01-04
fix(refs T30430): Add tiptap to externals

salisdemos:
We have to add `tiptap` to externals to avoid unexpected behaviors of `prosemirror` if `tiptap` or `prosemirror` is used in the target project as well. 

During testing/development the yarn.lock-File got updated. I cant say why, but I think it may be part of the fix.

The added tiptap-dependencies in https://github.com/demos-europe/demosplan-core/pull/387 where removed in the meanwhile. That may causes the errors. never the less I think we shouldn't re-add it there. So I tried this solution.

**Ticket** https://yaits.demos-deutschland.de/T30430


graupnerdemos (approved):
We're reaching the point with the externals where it's likely easier to keep an exclude list from the `dependencies` of the `package.json` and just use that one instead.

spiess-demos (changes requested):
I'd fix this within demosPlan.

salisdemos:
> I'd fix this within demosPlan.

Arg. Automerge was enabled

PR #56, opened 2023-01-05
Fix: Set a condition if the root element is empty WIP

elmasdemos:
Description: In assessment table in the page of original statement the console shows some errors. the problem could be that the root element is empty, so we need a condition to fix this problem. Anyway I couldnt test it yet, if it works this way


salisdemos (approved):
lgtm.

it only can make things better ;-)

PR #59, opened 2023-01-05
Revert "fix(refs T30430): Add tiptap to externals"

salisdemos:
Reverts demos-europe/demosplan-ui#55

This shall be fixed in demosplan-core


spiess-demos (approved):
thanks!

PR #61, opened 2023-01-06
Transpile tiptap esm modules

spiess-demos:
Ticket: https://yaits.demos-deutschland.de/T30430

To not run into conflicts between es6 and es5, the tiptap modules (that export esm bundles) must be included for babel-loader. This was already the case in the demosplan-core build but got overseen when moving DpEditor into demosplan-ui.

See https://stackoverflow.com/a/51860850/6234391

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## UNRELEASED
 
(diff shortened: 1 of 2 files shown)


salisdemos (approved):
Thanks so much for finding the reason.

PR #65, opened 2023-01-17
fix(refs T31010): Handle ES5/ES6 issues

salisdemos:
On some Pages we found strange behavior resolving some external Dependencies. We thought, they where already fixed by transpiling them during the buildstep, but it turned out that it wasn't enough.

here https://github.com/webpack/webpack/issues/2933 I found another solution on top, which I applied and which looks like it works.

You can test this in demosplan-core -> /einstellungen/plattform (as Support)

ATM there are still changes to make it localy testable.
(changes in package.json)

refs https://yaits.demos-deutschland.de/T31010

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,6 +2,7 @@ const path = require('path');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const { VueLoaderPlugin } = require('vue-loader');
 const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
+const DefinePlugin = require('webpack').DefinePlugin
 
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -55,6 +56,9 @@ const config = {
   plugins: [
     new MiniCssExtractPlugin(),
     new VueLoaderPlugin(),
+    new DefinePlugin({
```
spiess-demos, line 59: I do not understand what is achieved by DefinePlugin here.

salisdemos: Me neither. I hoped You got an Idea. It seems to fix the issue, but why - I can't say. :-/

salisdemos:
This was the wrong track

PR #66, opened 2023-01-18
fix(refs T31010): Don't use Translator directly in template

salisdemos:
it somehow causes an error of undefined Translator. I Think this is not the underlying issue, but changing this helps me to get closer to the real problem


elmasdemos (approved):
I think the problem with the trans keys will be in other vue components, too, because we used it this way often.
lgtm, code review

PR #67, opened 2023-01-18
chore: write Changelog

salisdemos:
Add changes to changelog hence they have been forgotten

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## UNRELEASED
 
-- Fix: Move Translator out of template
-- chore: Bump Dependenices
+- Fix: `undefined` Translator error in DpResizableImage (#66)
+- chore: Bump Dependenices (#63)
 
 ## v0.0.7 - 2023-01-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+- Fix: Move Translator out of template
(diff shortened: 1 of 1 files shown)


elmasdemos (approved):
thanks! lgtm, code review

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+- Fix: Move Translator out of template
```
hwiem, line 5: I find this too unspecific. Which template? Also: Why? What does it fix? (Of course we don't want to go into too much detail here, but a little context would be helpful)
I think it's a good idea to include the type (fix, chore, etc.) here, but we should agree on it and then use it consistently.

hwiem (changes requested):

hwiem:
Side note: Could you reference the PR(s) this changelog change refers to in the PR description? :pray:

CHANGELOG.md
spiess-demos: I think there is no need in mentioning any implementation detail. If this fixes a bug, it can be mentioned here.

hwiem: Yeah, something like "fix `undefined Translator` error in `componentName`" would do

salisdemos: True. changed in 97506fe0ffcecee2fe83ea1af4a4ebed56667983

spiess-demos (approved):
Lgtm!

hwiem (approved):

PR #68, opened 2023-01-19
refactor(refs T30969): move DpAccordion to components root

hwiem:
Moves DpAccordion from the `core` directory to the `components` directory (so we can remove the `core` directory later).


src/components/DpAccordion/DpAccordion.stories.mdx
```diff
@@ -0,0 +1,28 @@
+import { Meta } from '@storybook/addon-docs'
+import DpAccordion from './DpAccordion'
+
+<Meta
+    title="Components/Accordion"
+    component={DpAccordion}
+/>
+
+# Accordion
+
```
spiess-demos, line 10: Maybe we can add a paragraph on when to use the accordion? Exhaustive [information on this](https://design-system.service.gov.uk/components/accordion/) can be found in the GOV.UK design system. In [the documentation on the `Details` component](https://github.com/demos-europe/demosplan-ui/blob/main/src/components/DpDetails/DpDetails.stories.mdx#details) i already mentioned the differentiation of accordion vs. details.

hwiem: At first I had the idea of adding basic documentation for every component that I move from the `core` to the `components` folder, but it slows down the process of cleaning up the components and we actually have a separate ticket for the documentation (https://yaits.demos-deutschland.de/T30976) which @muellerdemos wants to work on, so I decided to just move the files here.

spiess-demos (approved):
lgtm!

PR #69, opened 2023-01-19
refactor(refs T30969): move DpButtonRow to components root

hwiem:
Moves DpButtonRow from the `core` directory to the `components` directory (so we can remove the `core` directory later).


src/components/core/DpEditor/DpLinkModal.vue
```diff
@@ -46,7 +46,7 @@
 </template>
 
 <script>
-import DpButtonRow from '../DpButtonRow'
+import DpButtonRow from '../../DpButtonRow/DpButtonRow'
```
spiess-demos, line 49: The longer i look at those paths, the more i want to [use aliases to access other modules](https://webpack.js.org/configuration/resolve/#resolvealias). However i guess this can be a completely separate PR.

hwiem: I was wondering about the paths, too. We could also import from the index file, would you consider that a better solution?

PR #70, opened 2023-01-19
refactor(refs T30969): move DpFlyout to components root

hwiem:
Moves DpFlyout from the `core` directory to the `components` directory (so we can remove the `core` directory later).


spiess-demos (approved):
lgtm!

PR #72, opened 2023-01-19
Use aliases for components, directives, mixins and utils

spiess-demos:
To not always have to write cumbersome path declarations like '../../component/component', new aliases for the components, directives, mixins and utils folders are introduced.

Also, this makes reorganizing components within folders a breeze.

Aliases start with `~~/` to differentiate the aliased modules from modules published by an organization (like '@demos-europe').


src/components/DpCheckbox/DpCheckbox.stories.mdx
```diff
@@ -1,9 +1,6 @@
 import { Canvas, Meta, Story } from '@storybook/addon-docs'
 import DpCheckbox from './DpCheckbox'
 
-<Meta
```
hwiem, line 4: Is this not needed?

hwiem (approved):
thanks, very nice!
just 1 question

salisdemos (approved):
looks good. Thanks for the improvement. Just the Question why You decided to use `~~` instead of `@` like in demosplan-core?

spiess-demos:
> Just the Question why You decided to use `~~` instead of `@` like in demosplan-core?

@hwiem and me discussed this yesterday. at first, we went with `@`, too. but as this character is also used as a prefix for [npm organizations](https://docs.npmjs.com/about-organization-scopes-and-packages), we found it less confusing to introduce a distinct character. As [resolve.alias takes precedence over other module resolutions](https://webpack.js.org/configuration/resolve/#resolvealias), it would not introduce any errors in using the `@`, but the visual distinction adds to readability, i think.

spiess-demos:
Superseeded by https://github.com/demos-europe/demosplan-ui/pull/324

PR #73, opened 2023-01-20
bug(refs T27479): fix bulk edit for documents

muellerdemos:
### Ticket: 
https://yaits.demos-deutschland.de/T27479

### Description:

- added a new bus instance to enable a better communication between the treelist components to set the selected state
- somehow there was a reactivity issue with the state drilling architecture underlying before hence we decided to use a bus to communicate the selected state
- to avoid deep nested child components select states to be passed from node to node to the upper level (which is also very computing heavy) we decided to pass the select state via a newly introduced bus instead of emitting events via vue components
- this solves the reactivity problem which was underlying before when selecting all checkboxes at once
- before the implementation it was not possible to bulk edit child nodes when selecting a parent node since only the parent node was recognized as selected by vue which is technically wrong

### Related PRs:
https://github.com/demos-europe/demosplan-core/pull/648

### How to review/test

Go to Verfahren -> Dokumente and select some documents. There should be a bulk edit header appear, where you can delete or publish the documents.


src/components/core/DpTreeList/DpTreeList.vue
```diff
@@ -81,6 +81,7 @@
 </template>
 
 <script>
+import bus from './utils/bus'
```
spiess-demos, line 84: Nitpick: alphabetic order.

src/components/core/DpTreeList/DpTreeListNode.vue
```diff
@@ -98,6 +98,7 @@
 </template>
 
 <script>
+import bus from './utils/bus'
```
spiess-demos, line 101: Nitpick_ alphabetic order

spiess-demos (changes requested):
I'd suggest to add an entry to the unreleased section of the CHANGELOG like "fix: emit correct selection in DpTreeList (#73)"

src/components/core/DpTreeList/DpTreeList.vue
muellerdemos: true, done in [7d116bb](https://github.com/demos-europe/demosplan-ui/pull/73/commits/7d116bb84e96528540a82036a9e9c258a2869af1)

src/components/core/DpTreeList/DpTreeListNode.vue
muellerdemos: [7d116bb](https://github.com/demos-europe/demosplan-ui/pull/73/commits/7d116bb84e96528540a82036a9e9c258a2869af1)

muellerdemos:
> I'd suggest to add an entry to the unreleased section of the CHANGELOG like "fix: emit correct selection in DpTreeList (#73)"

You are right we should have an entry in the changelog, hence added here [b0db042](https://github.com/demos-europe/demosplan-ui/pull/73/commits/b0db042292489bb9f625380e336d781f6315f9d2)

spiess-demos (approved):
lgtm! core review only, however.

PR #79, opened 2023-01-26
F t30977 no more permission queries in demosplan UI

ahmad-demos:
https://yaits.demos-deutschland.de/T30977

- remove hasPermission from DpEditor.
- remove hasPermission from DpObscure & editorObscure.


src/components/core/DpObscure.vue
```diff
@@ -1,20 +1,13 @@
...
 </template>
 
 <script>
 export default {
-  name: 'DpObscure',
-
-  computed: {
-    obscureClass () {
-      return (hasPermission('feature_obscure_text')) ? 'u-obscure' : ''
-    }
```
salisdemos, line 16: I fear this one is a bit tricky. 
If we simply remove the permission here, we loose the benefit of the check.
At the moment it behave like this: 
If the permission is true, the Text gets a black background. But if not, It comes along like a regular text.
After backing the class in, we loose the effect of displaying the regular text.
I think we have to add a prop for the permission here. But:
the changes in `demosplan-core` won't be simple. DpObscure is implemented in a twig-filter and is registered globaly because some text wich may uses dpObscure comes from the Database.

salisdemos (dismissed):

src/components/core/DpObscure.vue
spiess-demos: > If the permission is true, the Text gets a black background. But if not, It comes along like a regular text.
After backing the class in, we loose the effect of displaying the regular text.

That was my first thought as well. But under which circumstance would we want to show the obscured text at all? 

Within admin editor instances, users still see the obscured texts in non-edit-mode

image

as well as edit-mode:

image

When exporting statements, the whole text is passed through the demosplan/DemosPlanCoreBundle/Logic/EditorService.php:185, replacing everything with █ . for non-obscured exports, the existence of that class within the markup will have no effect as it is rendered in non-html contexts.

salisdemos: :thinking:  Ok. You may be right. But after this changes are installed in demosplan-core these Changes I would please You @ahmad-demos to tell the testers to have a closer look at that obscure-feature. :pray:

salisdemos (approved):
see discussion.

src/components/core/DpObscure.vue
spiess-demos: +1 for that!

PR #80, opened 2023-01-26
fix(refs:T31024): Space missing by copy from docx WIP

elmasdemos:
**Ticket**: https://yaits.demos-deutschland.de/T31024

**Description**:
- if you copy a text from a word docx into tiptap, some spaces are missing


src/components/core/DpEditor/DpEditor.vue
```diff
@@ -1153,6 +1153,9 @@ export default {
           // Strip img tags from pasted and dropped content
           returnContent = returnContent.replace(/<img.*?>/g, '')
 
+          // Strip line break spaces, white spaces and new lines
+          returnContent = returnContent.replace(/(&nbsp;|\r|\n)/gmi, '')
```
spiess-demos, line 1157: This looks like all those characters are replaced by the empty string, which leads to all spaces between words being dropped. I guess this is not what we want. I'd suggest to replace the characters with one space character.

```suggestion
          returnContent = returnContent.replace(/(&nbsp;|\r|\n)/gmi, ' ')
```

spiess-demos (changes requested):
Apart from the suggestion i think that the changes in package.json should not be part of this PR.

src/components/core/DpEditor/DpEditor.vue
elmasdemos: thank you, it thought i put a space in already

elmasdemos: done https://github.com/demos-europe/demosplan-ui/pull/80/commits/82f72fb092309951b759dd14d6ad171507ea40e1

hwiem:
Can we close this? The bug was fixed in https://github.com/demos-europe/demosplan-ui/pull/93

elmasdemos:
closed because fixed in https://github.com/demos-europe/demosplan-ui/pull/93

PR #76, opened 2023-01-23
refactor(refs T30969): move core components

hwiem:
Moves some components from `components/core` to `/components`:

- DpContextualHelp + documentation
- DpCheckboxGroup
- DpInlineNotification
- DpProgressBar
- DpPager + documentation
- DpSlidingPagination
- DpResettableInput + documentation
- DpSplitButton

Ticket: https://yaits.demos-deutschland.de/T30969

PR #71, opened 2023-01-19
F t30968 move demosplan utils to demosplan UI

ahmad-demos:
https://yaits.demos-deutschland.de/T30968

- Merge demosplan-utils with demosplan-ui.
-- lib
-- mixins
-- tests
-- utils

PR #81, opened 2023-01-26
Fix paths after failed merge

spiess-demos:
At the moment, **main** can't be build. As files have moved in both #76 and #71, resolving of conflicts was not happening as magically as it looked. This PR fixes these issues.

For the `addFormHiddenField` method, which has been moved to demosplan-core, we decided to duplicate it here (as long as "classic" forms have to be supported), because the only alternative approach that makes sense here would be to introduce some kind of hooks in the validation process, which is out of scope for dpValidate as of now.


elmasdemos (approved):
lgtm, thank you for the fix!

hwiem (approved):
lgtm + build is fixed, thanks!

PR #90, opened 2023-01-27
Bump all storybook modules to 6.16

spiess-demos:
As it is cumbersome to do this via GitHub/dependabot ui.


salisdemos (approved):
thanks for the shortcut

PR #93, opened 2023-02-07
fix(refs T31024): keep empty spaces

hwiem:
Ticket: https://yaits.demos-deutschland.de/T31024

- when copy/pasting from word, empty spaces are no longer removed
- another bug was fixed along the way (dynamically importing `EditorObscure` made it unavailable, so when testing in demosplan, the editor was not rendered)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,10 @@
 
 ## v0.0.9 - 2023-01-23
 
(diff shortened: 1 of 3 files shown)


salisdemos (approved):
lgtm.

But does it mean, that we can't use dynamic imports in demosplan-ui?

sakutademos (approved):
lgtm! Thx!

PR #95, opened 2023-02-10
Do not unregister boilerplates module 

spiess-demos:
It is not needed anymore, as the boilerplates store is registered in bundle entrypoints now.

Also, it caused a bug when an outer component was destroyed but remounted afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Do not unregister boilerplates store module beforeDestroy
+- ([#95](https://github.com/demos-europe/demosplan-ui/pull/95)) Do not unregister boilerplates store module beforeDestroy ([@spiess-demos](https://github.com/spiess-demos))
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,21 @@
 
 ## UNRELEASED
 
-## v0.0.9 - 2023-01-23
+### Fixed
+
+- Do not unregister boilerplates store module beforeDestroy
```
hwiem, line 7: In order to keep things consistent, I'd suggest to link the PR and the author here as in the entries below

hwiem (changes requested):
Code lgtm, one nitpick though

CHANGELOG.md
spiess-demos: done in 75c08c79442446578ed2b77d0236c270a9f6a533

hwiem (approved):

PR #100, opened 2023-02-15
Set correct value for `aria-live` in DpNotifyContainer

spiess-demos:
This PR sets the correct value for `aria-live` in DpNotifyContainer when document becomes visible. The logic was mistakenly inverted before.


graupnerdemos (approved):
Good catch, thanks!

PR #107, opened 2023-02-27
Make white and black themeable

spiess-demos:
Some projects will require to override black and white. So here we go.


ahmad-demos (approved):
lgtm, codeReview, thanks!

PR #109, opened 2023-02-28
Remove demosplan-utils dependency

spiess-demos:
As we [merged demosplan-utils into demosplan-ui](https://github.com/demos-europe/demosplan-ui/pull/71), it seems like we forgot to remove the dependency. This is done here.

The dependencies of demosplan-utils are added here.


elmasdemos (approved):
lgtm, code review

PR #114, opened 2023-03-03
Feature: make demosplan UI available in addons

muellerdemos:
### Description

- This config should be universally valid in our core applications and in addons
- In our core application the demosplan-ui will be transformed by webpack to a new vue instance: `var Vue = __webpack_require__`
- However in addons, webpack resolves the package differently. It is processed like this internally: `!function(e,t){ true?module.exports=t()`
- So in addons the config we use for our core applications throws undefined webpack module errors when not defining an explicit library name and type. When setting the library name and the type to var, webpack uses the library.name as internal identifier which makes demosplan-ui available in addons ultimately.
- There are still things which remain unclear. Like why webpack uses different approaches in addons and in core applications. To fully get rid of umd format we need to understand this process way deeper than we do atm because what is happening exactly when in the webpack build is essential for this.


webpack.config.js
```diff
@@ -32,7 +45,11 @@ const config = {
     libraryExport: 'default',
     clean: true
   },
-  externalsType: 'umd',
+  resolve: {
+    extensions: ['.js', '.vue'],
+    symlinks: false
+  },
+  externalsType: 'commonjs',
```
spiess-demos, line 52: Why is this not set based on `env.addon`?

webpack.config.js
```diff
@@ -22,6 +23,18 @@ const transpileNodeModules = [
   'tiptap-extensions',
 ].map(module => resolve('node_modules/' + module))
 
+const addonOutput = {
+  path: path.resolve(__dirname, "dist"),
+  publicPath: '',
```
spiess-demos, line 28: What is the reason for setting this to empty string? Isn't this the default?

package.json
```diff
@@ -74,6 +74,8 @@
     "prepack": "yarn build && yarn build-tokens",
     "storybook": "start-storybook -p 6006",
     "build": "yarn build:prod",
+    "postinstall:addon": "yarn build:addon && yarn build-tokens",
```
spiess-demos, line 77: When will this hook be fired? Aka. How is demosplan-ui consumed within addons?

webpack.config.js
muellerdemos: the externals can always be set to commonjs and it still works they do not need to be in umd format. Since I prefer a solution to avoind the usage of the umd format completely this was a good starting point to reduce all occurences of such

package.json
muellerdemos: This hook needs to be fired when installing addons right before creating the addon components

webpack.config.js
muellerdemos: Sorry, this was a leftover and is not needed at all, dropped here [b4b01c4](https://github.com/demos-europe/demosplan-ui/pull/114/commits/b4b01c40910d04787bf59cc9f7d3a47df0ef1ecf)

package.json
graupnerdemos: Does this need to be part of the installation process in demosplan-core then?

muellerdemos: Yes this was my intention so we can use addons right away after the installation is done

webpack.config.js
muellerdemos: I was wrong here, the public path needs to be set to an empty string to make this config work. reverted here: https://github.com/demos-europe/demosplan-ui/pull/114/commits/7f4356a73c080a250cb91303069143da088d359a

package.json
muellerdemos: Those hooks were dropped since the current config is now valid for our core applications and also for our addons. Unfortunately I could not get rid of umd yet but this is a starter to reduce the duplications of components in our addons.

webpack.config.js
```diff
@@ -25,14 +25,18 @@ const transpileNodeModules = [
 const config = {
   entry: resolve('./src/index.js'),
   output: {
-    path: path.resolve(__dirname, 'dist'),
+    path: path.resolve(__dirname, "dist"),
```
spiess-demos, line 28: Nitpicky suggestion:

```suggestion
    path: path.resolve(__dirname, 'dist'),
```

webpack.config.js
```diff
@@ -26,13 +26,17 @@ const config = {
   entry: resolve('./src/index.js'),
   output: {
     path: path.resolve(__dirname, 'dist'),
+    publicPath: '',
     filename: 'demosplan-ui.umd.js',
-    library: '@demos-europe/demosplan-ui',
-    libraryTarget: 'umd',
+    library: {
+      name: '_demos_europe_demosplan_ui',
```
graupnerdemos, line 32: I've seen custom library names starting with two underscores instead of one out there. I am however not sure if that is a convention or if I've just been lucky enough to only stumble upon this style.

graupnerdemos (approved):
Let's call it progress. It works for me.

webpack.config.js
muellerdemos: It seems common to use the double underscore start hence adjusted here [f72ac49](https://github.com/demos-europe/demosplan-ui/pull/114/commits/f72ac49ebb2ebf3371ec29df00059f4d21b9d000)

PR #127, opened 2023-03-16
Badges!

spiess-demos:
To document that we agreed on the usage of common-changelog, the badge is added in the README. As this one badge looks very lonesome, a license badge is destined to accompany it.

The choice of displaying the MIT license badge with a green color expresses a permissive license.


hwiem (approved):
nice!

PR #128, opened 2023-03-16
f_adjust_package_json_version_number

muellerdemos:
the latest version of demosplan-ui is 0.0.16 so should be adjusted in package.json accordingly


spiess-demos (approved):
lgtm.

PR #132, opened 2023-03-20
fix: Remove empty file

gruenbergerdemos:
With https://github.com/demos-europe/demosplan-ui/pull/104 DpBoilerplate was moved to core. This file was accidentally just emptied instead of deleted.


hwiem (approved):
lgtm

PR #133, opened 2023-03-20
fix(refs T31929): Make DpDatepicker loadable

salisdemos:
The library we are using ([a11y-datepicker](https://github.com/mathislucka/a11y-datepicker/)) was iomported as es module. That didn't works with the current setup, so we just have to load the cjs module and everything is fine.

**Ticket** https://yaits.demos-deutschland.de/T31929

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
     "vue-omnibox": "^0.3.7",
     "vue-sliding-pagination": "^1.3.2",
     "vuedraggable": "^2.24.3",
-    "vuex": "^3.6.2",
+    "vuex": "^3.6.2"
+  },
+  "devDependencies": {
     "@babel/core": "^7.20.2",
     "@babel/preset-env": "^7.20.2",
     "@storybook/addon-actions": "6.5.16",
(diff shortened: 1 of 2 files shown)


spiess-demos (commented):
Nitpick: i think this fix deserves a changelog entry.

salisdemos:
> Nitpick: i think this fix deserves a changelog entry.

Thanks for the Advice, but what should I write there?
```
BREAKING: Fix: DpDatepicker now gets loaded as cjs
```
?

spiess-demos:
> > Nitpick: i think this fix deserves a changelog entry.
> 
> Thanks for the Advice, but what should I write there?
> 
> ```
> BREAKING: Fix: DpDatepicker now gets loaded as cjs
> ```
> 
> ?

The "BREAKING" would be applied to the change of the build. The change you're applying is just the fix. However if both of them are within the same release i doubt that we should mention the fix at all...

hwiem (commented):
I think this can work as a quick fix, but we should still look for the underlying cause that I guess is related to changes in our webpack config?

salisdemos:
Yeah. I try to investigate a bit before merging this.

hwiem:
> For me, It looks like its related to the changes wich came with #114 But I am not sure. I tried to test this changes here with and without addons, but I am nur certain if my testing was fulfilling

So we will look for another solution in a later PR?

salisdemos:
> > For me, It looks like its related to the changes wich came with #114 But I am not sure. I tried to test this changes here with and without addons, but I am nur certain if my testing was fulfilling
> 
> So we will look for another solution in a later PR?

I would say so. or we leave it like it is but find the reason behind it. But for now I don't have any other solution.

hwiem (approved):

PR #138, opened 2023-03-29
Replace reserved chars in filenames

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T31971

As the tus endpoint breaks with 400 when uploading files with certain characters within them, the fileName is cleaned from those malfunctioning characters before adding it to the Uppy store (and eventually uploading them, as auto resume is activated).

The characters to be replaced were chosen based on a manual test that i conducted - while i could not produce a file containing a forward slash on my linux machine, all other characters have proven to break the endpoint. Maybe i missed a character, then it will have to be added to the list.

The fix actually only changes the file name in the outermost `file.name`, and within `file.meta.name`, whereas it leaves the original file name within `data.File.name` (which is an instance of the [File API](https://developer.mozilla.org/en-US/docs/Web/API/File) and also preserves the original name within its prototype). Somehow this leads to the unchanged file name being displayed correctly after uploading has finished. As such, there can be said that the bug has not been understood from a backend perspective.

The fix revealed another bug - file names containing a colon : now broke the frontend, because the colon is used as a separator in getFileInfo. As the conversion (from object to fileString and vice versa) is not needed but is a remainder from the Uppy refactoring (as far as i could understand it), it is dropped altogether, and the file object is directly passed into DpUploadFile.

### DIFF
--- a/src/components/core/DpUpload/DpUpload.vue
+++ b/src/components/core/DpUpload/DpUpload.vue
@@ -124,11 +124,6 @@ export default {
         '\\'
       ]
 
-      /*
-       * For reasons i could not figure out, the for loop was needed here
-       * to reliably access the item that is looped over. When using forEach,
-       * strangely it always held the value of currentFile.
-       */
       for (let i = 0; i < reservedCharacters.length; i++) {
         if (fileName.includes(reservedCharacters[i])) {
           fileName = fileName.replace(reservedCharacters[i], '_')


src/components/core/DpUpload/DpUpload.vue
```diff
@@ -95,6 +95,60 @@ export default {
...
+        '/',
+        ':',
+        '\\'
+      ]
+
+      /*
+       * For reasons i could not figure out, the for loop was needed here
+       * to reliably access the item that is looped over. When using forEach,
+       * strangely it always held the value of currentFile.
+       */
```
salisdemos, line 131: since its totally fine to use a for-loop, I suggest to drop the comment.

src/components/core/DpUpload/DpUpload.vue
```diff
@@ -95,6 +95,60 @@ export default {
...
+        }
+      }
+
+      if (fileName !== currentFile.name) {
+        return {
+          ...currentFile,
+          name: fileName,
+          meta: {
+            ...currentFile.meta,
+            name: fileName
```
salisdemos, line 144: what exactly happens with the meta-infos?

salisdemos (changes requested):

src/components/core/DpUpload/DpUpload.vue
spiess-demos: but in case somebody wants to optimize it, they'll be warned, if we leave the comment there.

src/components/core/DpUpload/DpUpload.vue
spiess-demos: > what exactly happens with the meta-infos?

I had to change the name also within them, otherwise it did not work out. I assume the backend logic someway uses the meta infos when actually creating the file.

src/components/core/DpUpload/DpUpload.vue
spiess-demos: ok, removed in 5c6ee110cf4c7be365761f85dae3744c624ede27

PR #139, opened 2023-03-29
fix: Revert "fix(refs T31929): Load cjs package instead of es"

salisdemos:
This reverts commit cca9f4219b62ceb786a74e5899b22f44ac01989a.


spiess-demos (approved):
lgtm (code review)

PR #146, opened 2023-04-05
chore: expose DpNotifyMessage and deprecate DpNotifyContainer

salisdemos:
`DpNotifyContainer` contains store logic which doesn't belong here. So we gonna import `DpNotifyMessage` directly in the Core


elmasdemos (approved):
lgtm, code review only

PR #148, opened 2023-04-05
Fix(refs T32170): refactor dynamic imports

muellerdemos:
Ticket: https://yaits.demos-deutschland.de/T32170

Dynamic imports are causing bugs every now and then, hence we decided is to refactor them and replace those with static imports to avoid future bugs. There is a tradeoff however since the bundle output will be slightly bigger this way but we decided it is still worth it, to adjust these imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## UNRELEASED
+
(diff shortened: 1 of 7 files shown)


salisdemos (approved):
legit

PR #149, opened 2023-04-05
chore: Remove usage of bus

salisdemos:
Since we are just in one single vue instance, we don't need to use another vue instance acting as bus. Its leaner to simply use the root event to communicate between nested componenents


spiess-demos (changes requested):
This is not the same bus as the core bus! It is a single usage solution designed only to improve performance on deeply nested treeLists when selecting many items at once.

salisdemos:
> This is not the same bus as the core bus! It is a single usage solution designed only to improve performance on deeply nested treeLists when selecting many items at once.

But since we don't have to manage the event trough different vue instances, we can use the $root-way. Creating another Vue instance without any needs is not a good idea

salisdemos:
> > > This is not the same bus as the core bus! It is a single usage solution designed only to improve performance on deeply nested treeLists when selecting many items at once.
> > 
> > 
> > But since we don't have to manage the event trough different vue instances, we can use the $root-way. Creating another Vue instance without any needs is not a good idea
> 
> I vaguely remember that using `this.$root.$on` did not work out, when implementing the fix. Also, when using several instances of DpTreeList within one root element, this won't work.

ok. Strange. That should do so. Or we have a problem with the way demosplan-ui uses Vue in the parent Environment.
I close this for now and will investigate later on (or hope someone else will)

PR #159, opened 2023-04-13
Tweak default offset for tooltip

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T29863

By specifying the offset via javascript (instead of css), a bug is fixed where the tooltip arrow was misplaced when a tooltip appeared directly in a corner of the parent element.

:warning: The related PR that changes the CSS accordingly has to be merged, after a new version of demosplan-ui has been released.


salisdemos (approved):
Thanks for fixing the changelog along

hwiem (approved):
lgtm

PR #175, opened 2023-04-24
fix(refs T32170): adjust layout of sliding dppager

muellerdemos:
https://yaits.demos-deutschland.de/T31974

- the pagination selection should be before the dropdown and should not have a term between the pagination and the dropdown.


elmasdemos (approved):
lgtm, code review

PR #178, opened 2023-04-24
chore(refs T32012): Prevent size from being 0

salisdemos:
in preparation for vue3 we have to prevent size from beeing able to be 0


elmasdemos (approved):
thanks for the fix. lgtm code review

PR #179, opened 2023-04-24
chore(refs T32012): Try a way to make slot fallbacks work again

salisdemos:
In demosplan-core the default slot in buttons didn't work within the vue3 migration. So we have to change it sowehow.
this is no beatiful solution but it looks like it works.


elmasdemos (approved):
lgtm, code review. thanks

PR #180, opened 2023-04-24
chore(refs T32012): Use native setter instead of vue.set

salisdemos:
in preparation for vue3. this does the same and is deprecated in vue3


ahmad-demos (approved):
lgtm, codeReview!

PR #181, opened 2023-04-24
chore(refs T32012): unFunctional Multiselect

salisdemos:
In vue3 functional components doesn't work. So we have to migrate them all to template-components.
Starting with dpMultiselect


elmasdemos (approved):
lgtm, code review. thanks for the affort

PR #195, opened 2023-05-03
chore(refs T19787, T19759): Rebuild template for selectPageItemCount

salisdemos:
As Part of the migration to Vue3 we move all functional components to "classic" Compontents. There changed an lot and for sake of simplicity and consistancy having a template is great.

This PR moves only the render function content to the template

### DIFF
--- a/src/components/core/DpDataTable/DpSelectPageItemCount.vue
+++ b/src/components/core/DpDataTable/DpSelectPageItemCount.vue
@@ -3,11 +3,13 @@
     <select
       id="item-count"
       class="o-form__control-select width-auto u-mr-0_25"
-      @change="e => $emit('change-count', parseInt(e.target.value))">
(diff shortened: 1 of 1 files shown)


elmasdemos (approved):
thank you. code review, lgtm

src/components/core/DpDataTable/DpSelectPageItemCount.vue
```diff
@@ -1,9 +1,28 @@
+<template>
+  <div>
+    <select
+      id="item-count"
```
spiess-demos, line 4: using a hardcoded id in a component that may have multiple instances within a view will lead to invalid html. this may be solved with a [composable like this one](https://github.com/demos-europe/diplanportal-frontend/blob/main/composables/util/useUniqueComponentId.ts).

PR #200, opened 2023-05-04
Generalize column selection tracking

spiess-demos:
- use localStorageKey to generate distinct Event Actions
- passing a Set as Event Name did not work, so it has to be converted to string first


salisdemos (approved):
Thanks

PR #201, opened 2023-05-04
Fix(refs T19787, T19759): Pass propper Prop 

salisdemos:
I forgot to commit that in the original PR and no one noticed.
This Change is necessary that the trans key get passed properly

And the missing Changelog entry of course.

see here github.com/demos-europe/demosplan-ui/pull/195


spiess-demos (approved):
lgtm!

PR #202, opened 2023-05-04
chore(refs T19787, T19759):  Rebuild template for DpSlidingPagination

salisdemos:
As Part of the migration to Vue3 we move all functional components to "classic" Compontents. There changed an lot and for sake of simplicity and consistancy having a template is great.

This PR moves only the render function content to the template.

With that The DpdataTableExtended now ueses the DpSlidingPagination and not the dependency directly so it benefits from the aria definitions.


muellerdemos (approved):
lgtm

PR #203, opened 2023-05-04
chore(refs T19787, T19759): Rebuild template for DpContextualHelp

salisdemos:
unFunctional Component and Move Template from Render to Template

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+### Refactor
+- ([#203](https://github.com/demos-europe/demosplan-ui/pull/203)) Move from Functional Component to Template in DpContextualHelp ([@salisdemos](https://github.com/salisdemos))
+
 ## v0.0.21 - 2023-05-02
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## UNRELEASED
 
+### Refactor
```
hwiem, line 5: I think this should be in the "Changed" category (if we want to follow [common changelog](https://common-changelog.org/) conventions), or, if we want to introduce a separate category for refactorings, it should be "Refactored"

salisdemos: Thanks. I never find the resource back. Will change that

hwiem (approved):
lgtm

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## UNRELEASED
 
 ### Changed
+- ([#203](https://github.com/demos-europe/demosplan-ui/pull/203)) Move from Functional Component to Template in DpContextualHelp ([@salisdemos](https://github.com/salisdemos))
```
spiess-demos, line 6: I'm unsure if we have to mention internal implementation changes like those in the changelog. As i understand it, a changelog should communicate changed behavior, to enable consumers adapt their usages.

salisdemos: Me neither. I thought its worth mentioning it, so one can find errors faster if something strange happens after the update

PR #206, opened 2023-05-05
chore(refs T19787, T19759): Don't use Static Ids

salisdemos:
Don't use Static Ids for components that can appear multiple times on one page.

as a follow op for https://github.com/demos-europe/demosplan-ui/pull/195

### DIFF
--- a/src/components/core/DpDataTable/DpSelectPageItemCount.vue
+++ b/src/components/core/DpDataTable/DpSelectPageItemCount.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <select
-      id="item-count"
+      :id="selectId"
       class="o-form__control-select width-auto u-mr-0_25"
       @change="e => $emit('changed-count', parseInt(e.target.value))">
       <option
@@ -20,6 +20,7 @@
 </template>
 
 <script>
+import { v4 as uuid } from 'uuid'
 export default {
   name: 'DpSelectPageItemCount',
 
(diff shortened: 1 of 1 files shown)


src/components/core/DpDataTable/DpSelectPageItemCount.vue
```diff
@@ -29,14 +30,20 @@ export default {
...
+     required: true
+    },
+
     pageCountOptions: {
       type: Array,
       required: true
     },
 
-    labelText: {
+    selectId: {
```
spiess-demos, line 43: Why do we need to set this dynamically? I'd make it a computed instead.

src/components/core/DpDataTable/DpSelectPageItemCount.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div>
     <select
-      id="item-count"
+      :id="selectId"
```
spiess-demos, line 4: Line 15 has to be adapted then, too.

spiess-demos (changes requested):

src/components/core/DpDataTable/DpSelectPageItemCount.vue
salisdemos: like so? e16ace514de60e25f9ed320cbc145926a783f5da

src/components/core/DpDataTable/DpSelectPageItemCount.vue
salisdemos: thanks for spotting that e16ace514de60e25f9ed320cbc145926a783f5da

spiess-demos (approved):
lgtm!

PR #211, opened 2023-05-08
Swap fontawesome with phosphor icons

spiess-demos:
Ticket: https://yaits.demos-deutschland.de/T30598

Paths of the MIT licensed phosphoricons are copied. As phosphoricons generally come with some whitespace around them, sizing needed to be adapted.

Once we go Vue3 with demosplan-ui, we can directly use [@phosphor-icons/vue](https://github.com/phosphor-icons/vue) which also seems to be tree shakeable.


salisdemos (approved):
Thanks a lot. I think we only can see if all icons are still fitting after merging and observing it

PR #212, opened 2023-05-09
Expose isExpanded in scoped slot

spiess-demos:
To be able to use the expanded state of DpFlyout right within the consuming component, it must be exposed to the scoped slot.


elmasdemos (approved):
lgtm, code review

PR #227, opened 2023-05-16
Correct changelog

spiess-demos:
Things got mixed up, so that the most "dangerous" change was hidden in a version where it did not happen.


ahmad-demos (approved):
lgtm, codeReview!

PR #234, opened 2023-05-22
bug(refs T32834): multiselect

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T32834

- Add prop for searchable
- Add event listener for select
- Show buttons for select all and unselect all only if option for multiple is true
- Order attributes
- Allow number as value


spiess-demos (approved):
lgtm (code review)

PR #236, opened 2023-05-22
Fix(T32849): DpTableHeader layout and slots issue

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T32849

**Description:** This PR fixes the issue with DpTableHeader. 

- change the order of the columns
- pass the slots into `DpTableHeader` correctly
- set `indeterminate` state using HTMLInputElement object's indeterminate property via JavaScript as it cannot be set using an HTML attribute (more info [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate))


src/components/core/DpDataTable/DpTableHeader.vue
```diff
@@ -2,56 +2,61 @@
...
+        :checked="checked" />
+    </th>
     <template v-for="(hf, idx) in headerFields">
       <dp-resizable-column
         v-if="isResizable"
         :is-last="headerFields.length === idx"
         :header-field="hf"
         :idx="idx">
-        {{ hf.label }}
+        <slot :name="`header-${hf.field}`">
```
elmasdemos, line 32: instead of this way, I would suggest to write it this way:
`<slot :header-field="hf.field"`>
 and change it in the files, too where it has the reference
 because the problem is, this way the code is not so readable after a few time

elmasdemos (changes requested):

src/components/core/DpDataTable/DpTableHeader.vue
sakutademos: you mean not readable in Vue 3? Would it be something like `<slot :[`header-${field}`]="hf.field">`? Because `<slot :header-field="hf.field">` dose not work

spiess-demos (approved):
Lgtm (code review).

src/components/core/DpDataTable/DpTableHeader.vue
spiess-demos: :[\`header-${field}\`]="hf.field"  sounds good! (if it works)

elmasdemos: I see now. lgtm thanks for the affort

elmasdemos (approved):

PR #237, opened 2023-05-23
bug(refs T3285): Fix multiselect

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T32855

- Add more types for prop value
- Bind props to slot, so they are available for parent
- Show selection buttons only if prop for controls is true
- Add all event listener

### DIFF
--- a/src/components/DpMultiselect/DpMultiselect.vue
+++ b/src/components/DpMultiselect/DpMultiselect.vue
@@ -1,6 +1,7 @@
 <template>
   <div>
     <vue-multiselect
+      :close-on-select="closeOnSelect"
       :deselect-group-label="deselectGroupLabel"
       :deselect-label="deselectLabel"
       :label="label"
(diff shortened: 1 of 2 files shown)


src/components/DpMultiselect/DpMultiselect.vue
```diff
@@ -26,31 +31,37 @@
       </template>
 
       <template v-slot:option="props">
```
spiess-demos, line 33: Can we also just expose option in the slots? Then, we would not have to adapt all useages.

```suggestion
      <template v-slot:option="{ option }">
```

gruenbergerdemos: I tried this, but then I couldn't access the remove function and it did not work (`@click="props.remove(props.option)"`.

spiess-demos (approved):
lgtm (code review)

PR #249, opened 2023-05-24
fix(refs: T32865): checkbox at wrong position and console errors

muellerdemos:
Ticket: https://yaits.demos-deutschland.de/T32865

- `this` is not valid in template and causes console errors hence has been removed
- On `mounted` can be problematic when working with `$refs` since refs are filled after the component mounts. Hence, to avoid a race condition like scenario, we can simply use another lifecycle hook here. Since `beforeUpdate` was documented here https://vuejs.org/guide/essentials/lifecycle.html#lifecycle-diagram as the hook to use when data changes, this has been implemented accordingly
- updated gitignore to not track irrelevant folders anymore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 !.storybook
 tokens/scss
 
+.idea
 node_modules/
 npm-debug.log*
 yarn-debug.log*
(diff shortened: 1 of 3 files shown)


gruenbergerdemos (approved):
lgtm and works!

PR #259, opened 2023-05-26
T21517 Add subtle button variant

spiess-demos:
Ticket: https://yaits.demos-deutschland.de/T21517

Apart from the existing button styles, a more subtle variant is needed for a long time. And here it is added in `DpButton`.

To reduce confusion, the default button variant, which went without a name before (as its styles were directly attached to the `btn` class) is named now (you may call it `solid` from now on).

This PR also adds the ability of passing icons to the button component, dropping the slot in advance.

For changed styling, also see https://github.com/demos-europe/demosplan-core/pull/1436

In the future there may be other button variants ("link" which may appear like a text link). This variant, however, focuses on the reduced level of importance and visibility.


src/components/DpButton/DpButton.vue
```diff
@@ -69,13 +69,14 @@ export default {
...
-     * Variants of the button include `outline` (`text` to be implemented).
+     * Variants of the button. Possible values are `solid`, `outline`, and `subtle`.
      * When not specified, the default style (white on colored background) is applied.
      */
     variant: {
       required: false,
       type: String,
-      default: ''
+      default: 'solid',
+      validator: (prop) => ['solid', 'outline', 'subtle'].includes(prop)
```
hwiem, line 79: While we're at it, should we also add a validator for the `color` prop?

spiess-demos: right! b8c70ff121e704f8a479cb9ace370e7eb1aa4615

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ### Removed
 - ([#255](https://github.com/demos-europe/demosplan-ui/pull/255)) **Breaking**: Remove DpRegisterFlyout ([@hwiem](https://github.com/hwiem))
+- ([#259](https://github.com/demos-europe/demosplan-ui/pull/259)) **Breaking**: DpButton do not support slots anymore. Use the `icon` prop instead. ([@spiess-demos](https://github.com/spiess-demos))
```
hwiem, line 7:
```suggestion
- ([#259](https://github.com/demos-europe/demosplan-ui/pull/259)) **Breaking**: DpButton does not support slots anymore. Use the `icon` prop instead. ([@spiess-demos](https://github.com/spiess-demos))
```

src/components/DpButton/DpButton.vue
```diff
@@ -50,7 +85,25 @@ export default {
...
+    icon: {
+      required: false,
+      type: String,
+      default: ''
+    },
+
+    /**
+     * Icon that will be placed after button text.
+     */
+    iconAfter: {
```
hwiem, line 99: I thought this would be a boolean, or is there a case in which we have an icon both before and after the button text?

src/components/DpButton/DpButton.vue
```diff
@@ -69,13 +122,14 @@ export default {
     },
 
     /**
-     * Variants of the button include `outline` (`text` to be implemented).
-     * When not specified, the default style (white on colored background) is applied.
+     * The button may have a variant of `solid`, `outline`, and `subtle`.
```
hwiem, line 125:
```suggestion
     * The button may have a variant of `solid`, `outline`, or `subtle`.
```

src/components/DpButton/DpButton.vue
spiess-demos: even if it is not a relevant use case to support two icons at the same time, we would need to define both the `icon` and `iconAfter` prop to render an icon after the text. So, in this case i'd opt for a little bit of magic to have a simpler component api.

src/components/DpButton/DpButton.vue
```diff
@@ -89,22 +143,55 @@ export default {
...
     },
 
     element () {
       return this.isButtonElement ? 'button' : 'a'
     },
 
+    /*
+     * Render a tooltip to show when hovering icon-only buttons.
+     */
+    iconTooltip () {
```
hwiem, line 158: If I set `iconAfter` and `hideText = true`, the tooltip will not be displayed. We could handle this by adding a condition `|| (iconAfter !== '' && hideText` to the `v-if="icon"` span, and adjusting the condition on the `v-if="iconAfter"` span:
`v-if="iconAfter && !hideText"`

hwiem (commented):
One final remark, rest lgtm

src/components/DpButton/DpButton.vue
spiess-demos: good catch! however i think it is enough to just show the tooltip on the iconAfter element: 880ed8149eadb540021fabbb272dd176b9da5696

the tooltip will be rendered twice, if the button is used with two icons but without text. but that is an absurd case anyways.

src/components/DpButton/DpButtonIcon.vue
```diff
@@ -0,0 +1,27 @@
+<template>
```
hwiem, line 1: - Is this needed as a separate component?
- Shouldn't `icon` be a required prop?

spiess-demos: > Is this needed as a separate component?

Not really, but i felt it reduces duplication a little bit. It is not exposed as a separate component, only used in DpButton.

> Shouldn't icon be a required prop?

No, as the `v-if="icon"` (hence, the decision if the icon should be displayed) is implemented here.

hwiem: But does it makes sense to implement it inside the component? If there is no icon, the component is not needed at all, therefore I'd make the `icon` prop required.

spiess-demos: i just wanted to reduce redundancy of the icon and iconAfter code. implementing the v-if inside the child component enables to have the icon as one liner in the parent component like so:

```vue
<dp-button-icon :icon="icon" />
```

However if you feel that this harms readability, i'd be happy to move the code back into DpButton and remove DpButtonIcon.

hwiem: Since the component is not exposed, I don't have a strong opinion on this (so we may leave it as is). But normally I'd say it's too implicit and makes it more difficult to understand what's going on

hwiem (approved):
lgtm!

src/components/DpButton/DpButtonIcon.vue
spiess-demos: removed it in 8667073c15e14ea8cf77079cac60f5e11a0488fa

PR #260, opened 2023-05-26
refactor(refs T30969): move components to components root, pt. III

hwiem:

Moves
- DpTreeList + related components
- DpDataTable + related components
- DpStickyElement

from `/components/core` to `/components`

### Tasks
- [x] revert moving DpSwitcher (should be removed entirely and replaced by DpToggle)
- [x] revert moving DpChangeStateAtDate (should be moved to demosplan-core)


spiess-demos (commented):
DpChangeStateAtDate is a very specific component, i don't know if we want it to be part of demosplan-ui, at least with that name...

hwiem:
> DpChangeStateAtDate is a very specific component, i don't know if we want it to be part of demosplan-ui, at least with that name...

I also had doubts about keeping it, so maybe we should just remove it?

hwiem:
> DpSwitcher is used only one time, in the (deprecated) fragments view. The generic component is DpToggle here. Or do you think we want DpSwitcher as a separate component?

I guess we'll have to adjust DpToggle a bit? But then we should replace/remove DpSwitcher, thanks for pointing it out!

spiess-demos:
> I guess we'll have to adjust DpToggle a bit? But then we should replace/remove DpSwitcher, thanks for pointing it out!

Which adjustments do you see here?

hwiem:
> > I guess we'll have to adjust DpToggle a bit? But then we should replace/remove DpSwitcher, thanks for pointing it out!
> 
> Which adjustments do you see here?

I think DpToggle cannot handle labels (like 'fragment' and 'statement')?

spiess-demos:
> I think DpToggle cannot handle labels (like 'fragment' and 'statement')?

I see... i did not think of DpToggle as a replacement for the actual implementation of DpSwitcher... DpSwitcher is just a very special UX solution in that place.

However maybe you're right and we can enhance DpToggle to have labels inside...

hwiem:
> > I think DpToggle cannot handle labels (like 'fragment' and 'statement')?
> 
> I see... i did not think of DpToggle as a replacement for the actual implementation of DpSwitcher... DpSwitcher is just a very special UX solution in that place.
> 
> However maybe you're right and we can enhance DpToggle to have labels inside...

I think it's a good idea to try to reduce the toggle patterns we use, it's mostly that I don't have a good idea how to implement it for the DpSwitcher use case, but we can also discuss it later. For now I'd revert moving DpSwitcher?

spiess-demos (approved):
lgtm! code review only.

PR #268, opened 2023-05-31
test: Add some Vue tests and a setup for them

salisdemos:
I restored the Vue tests that got lost during the great shift.

Now, if You run `yarn run test` not only the js files get tested, but some Vue files too.

I did not changed the tests itself.


jest.config.js
```diff
@@ -1,20 +1,66 @@
+/**
+ * (c) 2010-present DEMOS E-Partizipation GmbH.
+ *
+ * This file is part of the package demosplan,
+ * for more information see the license file.
+ *
+ * All rights reserved
+ */
+
```
hwiem, line 9: I believe this block should be removed (here and other files)?

jest.config.js
```diff
@@ -1,20 +1,66 @@
...
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+const path = require('path')
+
+console.dir('<rootDir>')
```
hwiem, line 13: (Why) Is this needed?

tests/__mocks__/dplan_api_map_options_admin.json
```diff
@@ -0,0 +1,46 @@
+{
```
hwiem, line 1: Where is this used? (Wondering the same for some other files in the mocks directory)

tests/prefixClass.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  it('should return a list of classes prefixed', () => {
+    expect(prefixClass('test foo bar')).toEqual('dp-test dp-foo dp-bar')
+  })
+
+  it('should return a quereyselector with prefixed classes', () => {
+    expect(prefixClass('#test .foo .bar')).toEqual('#test .dp-foo .dp-bar')
+    expect(prefixClass('[data-some-stuff="its all right"] > .a-class > label')).toEqual('[data-some-stuff="its all right"] > .dp-a-class > label')
+  })
+
+  it('should return the same sting if there are no points at the beginning of a word', () => {
```
hwiem, line 28:
```suggestion
  it('should return the same string if there are no points at the beginning of a word', () => {
```

hwiem (changes requested):

tests/__mocks__/dplan_api_map_options_admin.json
salisdemos: I can't tell. I just moved it. Cleaning up would be a next step.

jest.config.js
salisdemos: its not 4dcb065c73634144073bcb626a8e19d43b66cd07

jest/shallowMountWithGlobalMocks.js
```diff
@@ -0,0 +1,50 @@
+/**
```
hwiem, line 1: This block still needs to be removed

tests/__mocks__/dplan_api_map_options_admin.json
hwiem: I think we should make sure we need the files before adding them (apart from that, I agree that we can clean up things later)

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 - ([#269](https://github.com/demos-europe/demosplan-ui/pull/269)) DpIcon: new selector `landscape`, `portrait`, or `square`, based on icon proportions ([@spiess-demos](https://github.com/spiess-demos))
 - ([#270](https://github.com/demos-europe/demosplan-ui/pull/270)) DpIcon: new icons (ai, copy) ([@spiess-demos](https://github.com/spiess-demos))
+- ([#268](https://github.com/demos-europe/demosplan-ui/pull/268)) Tests: Add Testsetup for Vue Files and some Tests ([@salis-demos](https://github.com/salis-demos))
```
hwiem, line 8: don't we add the newest entry at the top of the section?

hwiem (changes requested):
The files `Filter.mock.js`, `layer_json.mock.js`, and `MapLayer.mock.js` are not used anywhere, so they should not be added to demosplan-ui

CHANGELOG.md
salisdemos: do we? We can though. I haven't thought about it.

salisdemos:
> The files `Filter.mock.js`, `layer_json.mock.js`, and `MapLayer.mock.js` are not used anywhere, so they should not be added to demosplan-ui

I don't know why i am so unconcentrated. Sorry!
Removed all the unused mocks 6afd8abf37d5559c161434ab39b5591834629071

CHANGELOG.md
salisdemos: changed in 3edbde43e14a994be1aa5c2bca4282688c7b255c

hwiem (approved):
Thanks for adding the tests :heart:

PR #269, opened 2023-05-31
Set helper class based on icon proportions

spiess-demos:
To be able to visually fine tune icon alignment, the icon now has a class of `landscape`, `portrait`, or `square`, based on the proportion of the icon path dimensions.


src/components/DpIcon/DpIcon.stories.mdx
```diff
@@ -252,6 +252,17 @@ Use as trigger for entering fullscreen mode along with [`compress`](#compress)
     </Story>
 </Canvas>
 
+### File
```
spiess-demos, line 255: This was just missing, i did not want to have another PR for it.

hwiem (approved):
lgtm

PR #275, opened 2023-06-01
refactor(T32651): Optimize DpDatatable Refactoring

salisdemos:
Some suggestions to make the especially Dpdraggable work (better) and some interaction behaviour

We need to configure a handle, so not the whole row is clickable but only the click-handle

### DIFF
--- a/src/components/core/DpDataTable/DpDataTable.vue
+++ b/src/components/core/DpDataTable/DpDataTable.vue
@@ -4,8 +4,8 @@
       ref="tableEl"
       :class="tableClass">
       <colgroup
-          ref="colgroup"
-          v-if="headerFields.filter(field => field.colClass).length > 0">
+        ref="colgroup"
+        v-if="headerFields.filter((field) => field.colClass).length > 0">
         <col v-if="isDraggable || isSelectable" />
         <col v-for="field in headerFields" :class="field.colClass" />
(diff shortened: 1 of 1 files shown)


src/components/core/DpDataTable/DpDataTable.vue
```diff
@@ -36,68 +36,21 @@
...
-                v-bind="item" />
-            </template>
-            <template v-slot:flyout="item">
-              <slot
-                name="flyout"
-                v-bind="item" />
-            </template>
-          </dp-table-row>
-
-          <!-- DpTableRowExpanded -->
```
sakutademos, line 77: Thanks a lot for your support! I noticed a problem when using the` <component> ` to define the tag (either `dp-draggable` or `tbody`). The `dp-draggable `component does not work correctly with the expanded functionality when the expanded row is inside the `DpDraggable` component.

There is no `<!-- DpTableRowExpanded --> `part in your code. If I add expanded row to the `<component>` below the `<dp-table-row>`, the draggable functionality does not work even when I add `v-if="!isDraggable"` to the expanded row.

sakutademos: The previous version worked well because the expanded row was only in the `tbody` section.

PR #278, opened 2023-06-02
refactor(T32651): Optimize DpDatatable Refactoring

salisdemos:
The main change here is adding the class to the drag-handle.

The other changes are just removing an unneccessary template tag and linebreaks


salisdemos:
this on is obsolet

PR #290, opened 2023-06-12
bug(refs T33153): Add missing props for multiselect

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T33153

Add some missing props to wrapper component to pass to vue-multiselect. `custom-label` needs to have `undefined` as default, because otherwise we get an empty label if the parent didn't give one.

### DIFF
--- a/src/components/DpMultiselect/DpMultiselect.vue
+++ b/src/components/DpMultiselect/DpMultiselect.vue
@@ -2,15 +2,22 @@
   <div>
     <vue-multiselect
       :close-on-select="closeOnSelect"
+      :custom-label="customLabel"
+      :data-cy="dataCy"
       :deselect-group-label="deselectGroupLabel"
       :deselect-label="deselectLabel"
+      :disabled="disabled"
       :group-label="groupLabel"
       :group-select="groupSelect"
       :group-values="groupValues"
+      :id="id"
       :label="label"
+      :loading="loading"
+      :max-height="maxHeight"
       :multiple="multiple"
+      :name="name"
       :options="options"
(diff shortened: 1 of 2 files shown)


src/components/DpMultiselect/DpMultiselect.vue
```diff
@@ -2,12 +2,22 @@
   <div>
     <vue-multiselect
       :close-on-select="closeOnSelect"
+      v-bind="conditionalProps"
+      :data-cy="dataCy"
```
spiess-demos, line 6: I wonder if it would enhance readability + is a recognized pattern to move other prop definitions to v-bind, too. Something like...

```suggestion
      v-bind="{
        ...conditionalProps,
        dataCy,
        deselectGroupLabel,
        deselectLabel,
        disabled,
        groupLabel,
        //...
      }"
```

However idk if this adds to code quality at all.

sakutademos (approved):
lgtm + works!

src/components/DpMultiselect/DpMultiselect.vue
```diff
@@ -152,6 +230,12 @@ export default {
       default: ''
     },
 
+    tagPlaceholder: {
+      type: String,
+      required: false,
+      default: () => Translator.trans('tag.create')
```
spiess-demos, line 236: Not a blocker right now (it is merged anyways) but didn't we want to drop Translator in demosplan-ui?

salisdemos: But we still have no concept how to handle things like that. so unfortunately we have to do it this way until we have a technical solution

PR #292, opened 2023-06-13
Chore: Clean up and simplify tableRow condition

salisdemos:
in Vue3 having a v-if on the same element as a v-for causes trouble. And this Changes should not do any harm in vue2


sakutademos (approved):
small nitpick, otherwise lgtm! Thx

PR #294, opened 2023-06-14
Prepare Tailwind config to be used within demosplan-core

spiess-demos:
This PR changes the Tailwind config so that it includes "Preflight", the opinionated normalize of Tailwind. So, we can drop normalize in demosplan-core. 

Also, it adds `yarn build:css` to the prepack steps, so that a complete tailwind build is always included in the package.

Lastly, it changes some util classes to be of the tailwind syntax, in preparation of them being dropped from demosplan-core.

:warning: As the utility classes are already changed within the components, https://github.com/demos-europe/demosplan-core/pull/1521 must also be merged.


tailwind.config.js
```diff
@@ -1,15 +1,33 @@
...
   safelist: [
     {
       pattern: /./// Disable purging https://github.com/tailwindlabs/tailwindcss/discussions/6557#discussioncomment-1838214
     }
   ],
   theme: {
-    extend: {}
+    extend: {
+      flexShrink: {
+        2: '2'
```
salisdemos, line 29: out of curiosity: Why that?

salisdemos (approved):
Just one Question, not a concern.

tailwind.config.js
spiess-demos: out of the box, [tailwind supports flex-shrink 0 or 1](https://tailwindcss.com/docs/flex-shrink). However in one place in demosplan we use flex-shrink: 2, and this is the official way of [extending tailwind utility classes](https://tailwindcss.com/docs/plugins#extending-the-configuration).

PR #304, opened 2023-06-16
fix: empty string is invalid for "variant" prop

spiess-demos:
this went unnoticed when implementing the subtle button...


hwiem (approved):
lgtm, thanks!

PR #314, opened 2023-06-19
chore: Use Barrel Export to simplify index.js

salisdemos:
Instead of importing and exporting every Item explicitly, we can simply use a so called barrel export.


gruenbergerdemos (approved):
nice, lgtm!

hwiem (approved):
Nice, thanks!

spiess-demos (commented):
Does this work in all circumstances? those are: tests, addons, storybook.

Before, we exposed modules via one huge **default** export (object). Now we forward the index files inside the directories that are barrel exported, instead. Those are named exports instead of default exports.

I do not know if this has a negative impact at all, just want to make sure that nothing is overlooked here. I also like the cleanliness here for sure.

salisdemos:
> Does this work in all circumstances? those are: tests, addons, storybook.
> 
> Before, we exposed modules via one huge **default** export (object). Now we forward the index files inside the directories that are barrel exported, instead. Those are named exports instead of default exports.
> 
> I do not know if this has a negative impact at all, just want to make sure that nothing is overlooked here. I also like the cleanliness here for sure.

As far as I can see, Its all fine. Storybook still works. Tests pass. tested it in demosplan-core. Couldn't find anything that its not treeshakeable.... And if we find something, we still can revert it

spiess-demos (approved):
lgtm! code review only.

PR #318, opened 2023-06-20
Feat(T33336): add externalApi method for external requests

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33336

**Description:** Add an additional method called `externalApi `to make an external API callswithout default headers.

- `Content-Type` header should be omitted in calls to some map services (axios adds this header automatically)
- delete the header `Content-Type` directly and then add it back after

### Linked PRs 
- https://github.com/demos-europe/demosplan-core/pull/1581


gruenbergerdemos (approved):
lgtm

PR #319, opened 2023-06-22
bug(refs T33411): Add prop for data-cy on its own

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T33411

We need to give the data-cy prop on its own, because otherwise it will be added as `dataCy` instead of `data-cy`.


spiess-demos (approved):
Lgtm. However what is happening here? If passed via v-bind, the camelCase does not translate to dashed case but case is ignored altogether?

gruenbergerdemos:
I think it is because the child component (vue-multiselect) has no prop `dataCy` and that's why it is not translated to dashed case, but just passed on as it is.

PR #291, opened 2023-06-13
chore(refs T30981) Typescript support demosplan UI

ahmad-demos:
https://yaits.demos-deutschland.de/T30981

- WIP: Add Typescript Config.
- WIP: Add DpRadio and DpButton component as Typescript.

Test:
In Core:
- go to `package.json:44`, then add `"@demos-europe/demosplan-ui": "github:demos-europe/demosplan-ui#200e416bbbb47eda2e9e442ada918d3ad3fc5661",`
- delete `node_modules` folder.
- then `yarn install`.
- in Terminal in phpStorm `cd node_modules/@demos-europe/demosplan-ui` -> `yarn install` -> `yarn build:prod` -> `yarn build-tokens`.
- at the end `dp shell jsdev:project`

:information_source: 
We have to use `.vue` when we import Vue-Components in TS-files.
https://stackoverflow.com/questions/52880039/how-to-remove-vue-extension-from-imports-when-using-typescript-in-vue-js

PR #324, opened 2023-06-22
Use internal alias "@/"

spiess-demos:
To not always have to write cumbersome path declarations like '../../component/component', a new alias "@/" is introduced, which simply points to the src folder. In the [predecessor of this PR, i suggested aliases](https://github.com/demos-europe/demosplan-ui/pull/72) for different folders (like components, mixins, and so on) but this seemed to be too much overhead at hindsight.

Despite the discussion in the other PR on why to chose `~~`, i realised that `@` is more common, so it is used as alias char here.

When trying to get jest, storybook and the webpack build up and running without errors, it turned out that storybook did not work well with the way we (default) export the vue components by themselves - whenever a component imported another component by alias, storybook could not render it anymore. Using index files within the component folders, which expose both a default and a named export, mitigated the issue. Also, index files in component folders seem to be a common pattern in libraries.

:factory_worker: I stopped moving everything to index files after DpContextualHelp because i first wanted to get feedback from the team. I guess it is best to finalize everything in a followup PR.

A few other thing this PR changes:

- move /shared/translations to new top level folder "locale"
- inline wrapItemIntoArray fn because it caused a ts error

Related PRs:

- Waits for https://github.com/demos-europe/demosplan-ui/pull/291
- Supersedes https://github.com/demos-europe/demosplan-ui/pull/72


spiess-demos:
Superseeded by https://github.com/demos-europe/demosplan-ui/pull/513

PR #325, opened 2023-06-26
chore (refs T30980): remove some component class, to pass it from out…

ahmad-demos:

- remove some component class, to pass it from outside.

https://github.com/demos-europe/demosplan-core/pull/1604


salisdemos (approved):

salisdemos:
Too late but please add a changelog-Entry.

PR #328, opened 2023-06-27
chore (refs T30980): Remove flex class from DpInlineNotification.

ahmad-demos:

- Remove flex class from DpInlineNotification.


spiess-demos (changes requested):
I think removing the `flex` utility class breaks the component. Which problem do you solve by doing so?

PR #333, opened 2023-06-27
Enhance DpButton and DpIcon components

spiess-demos:
This PR adds the "rounded" prop to buttons by using a tailwind utility class.

It also adds the arrow-down and arrow-up icons to DpIcon.


hwiem (approved):
lgtm, code review only

PR #335, opened 2023-06-29
chore (refs T30980): remove spacing class.

ahmad-demos:

- remove spacing class.


spiess-demos (approved):
lgtm! (code review only)

PR #343, opened 2023-07-03
refactor: Build demosplan-ui as module

salisdemos:


webpack.config.mjs
```diff
@@ -0,0 +1,106 @@
+import path from 'path'
+import MiniCssExtractPlugin from 'mini-css-extract-plugin'
+import { VueLoaderPlugin } from 'vue-loader'
+import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer'
+import { fileURLToPath } from 'url'
+
+function resolve (dir) {
+    return path.join(path.dirname(fileURLToPath(import.meta.url)), dir.replace('./', ''))
+}
```
salisdemos, line 9: for some reason git didn't manage to display the renaming and changes at once...

The `resolve` method changed

webpack.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+        publicPath: '',
+        filename: 'demosplan-ui.mjs',
+        library: {
+            // name: '__demos_europe_demosplan_ui',
+            type: 'module'
+        },
+        libraryExport: 'default',
+        libraryTarget: 'module',
+        clean: true
+    },
```
salisdemos, line 43: here are some changes

webpack.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+        new MiniCssExtractPlugin(),
+        new VueLoaderPlugin(),
+    ],
+    module: {
+        rules: [
+            {
+                test: /\.m?js/,
+                resolve: {
+                    fullySpecified: false
+                }
```
salisdemos, line 58: new

webpack.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+            },
+            {
+                test: /\.(eot|svg|ttf|woff|woff2|png|jpg|gif)$/i,
+                type: 'asset',
+            },
+        ],
+    },
+};
+
+export default function config() {
```
salisdemos, line 94: changed from `module.exports`

webpack.config.mjs
salisdemos, line 1: for some reason git didn't manage to display the renaming and changes at once...
So I tried to comment where I changed something. I hope I found all spots

src/index.js
```diff
@@ -6,3 +14,13 @@ export * from './mixins/'
...
+
+export default {
+  ...shared,
+  ...directives,
+  ...utils,
+  ...components,
+  ...mixins,
+  ...lib,
+  ...validation
+}
```
salisdemos, line 26: I don't understand why, but for some reason we need the default export AND the named Exports. I would love to understand that. I think then it should be easy to resolve it

spiess-demos: I've come across the same issues/fix within the [alias draft PR](https://github.com/demos-europe/demosplan-ui/pull/324). As i've seen the same pattern (having default exports besides named exports): maybe it's ok to do so.

spiess-demos: Here: https://github.com/demos-europe/demosplan-ui/pull/324/files#diff-6607fbfb793b10562651564bfa45bd5b8be08e5456b9a0b085341c8984a44e06R84-R89

eslintrc.js
```diff
@@ -0,0 +1,135 @@
...
+
+module.exports = {
+  root: true,
+  parserOptions: {
+    parser: '@babel/eslint-parser',
+    ecmaVersion: 6,
+    sourceType: "module",
+    ecmaFeatures: {
+      jsx: true
+    }
```
muellerdemos, line 23: i don't think we need the jsx support here since we are not using the syntax extension. From my perspective it is beneficial to keep the config as small as possible.

webpack.config.mjs
```diff
@@ -0,0 +1,106 @@
...
+    experiments: {
+        outputModule: true,
+    },
+    output: {
+        path: path.resolve('dist'),
+        publicPath: '',
+        filename: 'demosplan-ui.mjs',
+        library: {
+            // name: '__demos_europe_demosplan_ui',
+            type: 'module'
```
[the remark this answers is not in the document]
muellerdemos: I remember it was troublesome finding a config that is working in dplan and in our addons at the same time and I think I have tried something similar without much success. Have you tried already if this config is also working for both cases?

muellerdemos (commented):
The changes look reasonable to me so far. I've stumbled upon some module configs lately and I also wonder if we actually want to have the raw vue files also in our npm package. I've found out that more often than not only the bundled files are included in npm packages. This could be tweaked in the package.json under the `files` key

eslintrc.js
salisdemos: true. May be a leftover when I tried to get TS running.

webpack.config.mjs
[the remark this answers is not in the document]
salisdemos: Thanks for pointing that out. Didn't check that. Is it sufficent for testing to build the assets in an addon? Or should I install it for testing?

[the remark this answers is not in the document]
muellerdemos: The problem was not visible during build time but only in the browser. Unfortunately I think the way to go is to test this with an addon installed.

salisdemos:
I am not sure if this even was the right approach. And since it lay around quite a while, I would have to spend a lot of time to understand what my past Me did here. So if this is still the goal it would be easier to start again.

PR #358, opened 2023-07-07
ci(refs T33032): add jenkinsfile

shumyachkin:
adds jenkins pipeline for running jest tests 
example of the test result can be checked in jenkins


Jenkinsfile
```diff
@@ -0,0 +1,19 @@
+pipeline {
+    agent {label 'docker'}
+    options {
+        timeout(time: 1, unit: 'HOURS')
```
salisdemos, line 4: does the timeout has to be 1 hour?
I hope the tests run through way under 5 minutes.

shumyachkin: Good point, let's set it to 10 minutes first.

salisdemos (commented):
I don't feel in the position to approve this. That just would be formal

PR #361, opened 2023-07-07
Tweak Tailwind config with utility tokens

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T27596

 image
This PR adds a few more design tokens to demosplan-ui and configures Tailwind to use them.

:warning: **After releasing this, we must adapt scss paths within demosplan-core!**

- export tokens as js module to have them digestible by tailwind
- move json to tokens/src, move compiled scss and js to a new tokens/dist, to have a neat tokens folder
- add breakpoints and zIndex tokens, heavily inspired by demosplan-core but better and with naming partly adapted to tw
- add bc compatible scss variable names for boxShadow + rounded
- add glob to dependencies, it was missing before
- configure Tailwind with screens, borderRadius, boxShadow, and zIndex tokens, to be able to replace current demosplan-core utilities


tokens/src/zIndex.json
```diff
@@ -0,0 +1,36 @@
+{
+  "z-index": {
+    "below-zero": {
+      "$description": "Stack an element behind elements that do not have a stacking order applied.",
+      "value": "-1"
+    },
+    "zero": {
+      "$description": "Stack an element behind elements that do not have a stacking order applied.",
```
salisdemos, line 8: is it correct, that they both have the same description?

tokens/src/zIndex.json
```diff
@@ -0,0 +1,36 @@
...
+      "value": "2000"
+    },
+    "flyout": {
+      "$description": "Use for flyout components.",
+      "value": "2100"
+    },
+    "fixed": {
+      "$description": "Use for fixed or sticky elements and fullscreen elements.",
+      "value": "2200"
+    },
```
salisdemos, line 26: what happens if we have fixed and fullscreen at the same time? Don't we have to precise here?

tokens/src/zIndex.json
```diff
@@ -0,0 +1,36 @@
...
+    "modal": {
+      "$description": "Use for modals. Modals should be the topmost elements.",
+      "value": "2300"
+    },
+    "ultimate": {
+      "$description": "Use sparingly for elements that should even trump modals.",
```
salisdemos, line 32: :-D

salisdemos (changes requested):

tailwind.config.js
```diff
@@ -1,7 +1,36 @@
 const plugin = require('tailwindcss/plugin')
 
+const borderRadius = Object.fromEntries(
+  Object.values(require('./tokens/dist/js/rounded').rounded)
+    // Filter out aliases here because they are only included for backward compatibility with Scss
+    .filter(({ original}) => original.value.includes('{') === false)
```
hwiem, line 6:
```suggestion
    .filter(({ original }) => original.value.includes('{') === false)
```

tailwind.config.js
```diff
@@ -1,7 +1,36 @@
...
+    .map(({ attributes, value }) => [attributes.type, value])
+)
+
+const boxShadow = Object.fromEntries(
+  Object.values(require('./tokens/dist/js/boxShadow')['box-shadow'])
+    // Filter out aliases here because they are only included for backward compatibility with Scss
+    .filter(({ original}) => original.value.includes('{') === false)
```
hwiem, line 13:
```suggestion
    .filter(({ original }) => original.value.includes('{') === false)
```

tailwind.config.js
```diff
@@ -1,7 +1,36 @@
 const plugin = require('tailwindcss/plugin')
 
+const borderRadius = Object.fromEntries(
+  Object.values(require('./tokens/dist/js/rounded').rounded)
+    // Filter out aliases here because they are only included for backward compatibility with Scss
+    .filter(({ original}) => original.value.includes('{') === false)
+    .map(({ attributes, value }) => [attributes.type, value])
+)
+
+const boxShadow = Object.fromEntries(
```
hwiem, line 10: Nitpick: We could extract this code into a separate function to reduce code duplication

tokens/src/rounded.json
```diff
@@ -2,28 +2,43 @@
...
       "value": "4px",
       "themeable": true
     },
-    "large": {
-      "$description": "Large - The biggest radius. Use for boxes that purposefully should appear rounded - plays nice with, rounded elements inside it that are rounded via {rounded.small}.",
+    "medium": {
+      "value": "{rounded.md}"
+    },
+    "lg": {
+      "$description": "Large - The biggest radius. Use for boxes that purposefully should appear rounded - plays nice with, rounded elements inside it that are rounded via {rounded.default}.",
```
hwiem, line 32:
```suggestion
      "$description": "Large - The biggest radius. Use for boxes that purposefully should appear rounded - plays nice with rounded elements inside it that are rounded via {rounded.default}.",
```

tokens/tokens.zIndex.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+$dp-z-index-modal
+$dp-z-index-fixed
+$dp-z-index-flyout
+$dp-z-index-tooltip
+$dp-z-index-above-zero
+$dp-z-index-zero
+$dp-z-index-below-zero
+```
+
+## How to use Breakpoint tokens
```
hwiem, line 29:
```suggestion
## How to use z-index tokens
```

tokens/tokens.zIndex.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Then, tokens can be used within Scss code:
+
+```
+@media all and (min-width: $dp-breakpoints-lg) {
+  .some-selector {
+    display: block;
+  }
+}
```
hwiem, line 46: I think this is the wrong example

tokens/tokens.breakpoints.stories.mdx
```diff
@@ -0,0 +1,55 @@
...
+    display: block;
+  }
+}
+```
+
+
+### Tailwind utilities
+
+As Tailwind is configured with these breakpoints, too, they can be used as modifiers for tailwind utility classes.
+They replace the z-index classes mentioned within the Tailwind documentation.
```
hwiem, line 49: Should this be here?

tokens/tokens.boxShadow.stories.mdx
```diff
@@ -11,24 +11,20 @@ import { Meta } from '@storybook/addon-docs'
...
-$dp-box-shadow-2
-$dp-box-shadow-3
-$dp-box-shadow-4
+$dp-box-shadow-xl
+$dp-box-shadow-lg
+$dp-box-shadow-md
+$dp-box-shadow
 ```
 
+The `$dp-box-shadow-1` to `$dp-box-shadow-4` still work, but will be removed eventually.
```
hwiem, line 27: Should this be mentioned in the changelog entry?

tokens/src/boxShadow.json
```diff
@@ -1,24 +1,36 @@
...
   "box-shadow": {
-    "1": {
-      "$description": "Smallest elevation. Use for Cards + Buttons.",
+    "DEFAULT": {
+      "$description": "Default elevation. Use for Cards + Buttons.",
       "value": "0 1px 4px rgb(0 0 0 / 22%), 0 1px 2px rgb(0 0 0 / 26%)",
       "themeable": true
     },
-    "2": {
+    "1": {
```
hwiem, line 8: Can we add a `deprecated` comment here (and below)?

hwiem (changes requested):

tailwind.config.js
spiess-demos: right! 5a845e4f8f9157113e7c62e3c2ebad5596413bb6

tokens/src/zIndex.json
spiess-demos: The z-indexes represented here are reusable within any [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context). So, in the scenario of a fullscreen view that contains a fixed header toolbar, we could drop [an `isolate` TW utility class](https://tailwindcss.com/docs/isolation) onto the fullscreen element, and every child element within the fullscreen is magically stacked within that new context.

I've found some occurrences of `z-index: 0;` being (mis)used to establish a new stacking context within demosplan, these may be refactored to use `isolate`, too, to clearly communicate the intention.

Another thought with the small number of z-indexes: i see them as rough guidance on where to place an element. If an element needs a z-index that is "just below the other stuff of that element", say, the arrow of a flyout, the outer element can be positioned by these global z-indexes, while the inner elements can be positioned using a new stacking context.

tokens/tokens.boxShadow.stories.mdx
spiess-demos: yes. c3b79a2ceb30ff558f8fffc14d596a2683c0cc97

tokens/tokens.breakpoints.stories.mdx
spiess-demos: no. 93381674f64406a54ce690c36fbfa68141c8f0e3

tokens/src/zIndex.json
spiess-demos: You're right with laughing about this helpless begging... i hopefully made it clearer how to use those: 294edaf953a9fdede08d32d64eef92c7360bdfe5

tokens/src/zIndex.json
spiess-demos: no. 294edaf953a9fdede08d32d64eef92c7360bdfe5

hwiem (approved):
lgtm!

PR #362, opened 2023-07-10
chore: (refs T33353) replace linebreaks to blanks during paste

mussbach:
This PR reapplies a fix that was made in core in https://github.com/demos-europe/demosplan-core/pull/1657

**How to review/test**

Paste Text from Word (not office365) into some editor (e.g. "Textbausteine Anlegen"). All blank spaces should be preserved.

As a Dev, this can't be tested, only reviewed since there is no real word on linux.


salisdemos (changes requested):
I suggest to Close this one. 
The Bug that was fixed in the core repos (older release) wasn't reproducable in current main (demosplan-core). 
that fix was introduced only because applying https://github.com/demos-europe/demosplan-ui/pull/93 to the core didn't work out.

PR #363, opened 2023-07-12
Fix: Add a default export to index.js

salisdemos:
For some reasons we need both. Named and default exports. This should fix the Bugs wich leaded to the role back of 1.8. 

since I ordered the entries alphabetically, the changes look more than they are.

Only the second and third block is new. The upper block is just sorted


spiess-demos (approved):
lgtm from a code-review perspective.

PR #364, opened 2023-07-12
feat(refs T12070): Make IconSize a prop for DpButton

salisdemos:
To be more flexible when using the DpButton Component for clickable Icons, we have to make use of the different icon sizes which DpIcon allready provides.

### DIFF
--- a/src/components/DpButton/DpButton.stories.mdx
+++ b/src/components/DpButton/DpButton.stories.mdx
@@ -23,7 +23,11 @@ import DpButton from './DpButton'
     variant: {
       options: ['solid', 'outline', 'subtle'],
       control: { type: 'select' }
-    }
+    },
+    iconSize: {
+          options: ['small', 'medium', 'large'],
+          control: { type: 'select' }
+        }
   }}
   title="Components/Button"
   component={DpButton} />
--- a/src/components/DpButton/DpButton.vue
+++ b/src/components/DpButton/DpButton.vue
@@ -11,7 +11,7 @@
       v-if="icon"
       aria-hidden="true"
       :icon="icon"
-      size="small" />
+      :size="iconSize" />
     <span
       :class="{'hide-visually': hideText}"
       v-text="text" />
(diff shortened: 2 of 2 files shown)


src/components/DpButton/DpButton.vue
```diff
@@ -11,7 +11,7 @@
       v-if="icon"
       aria-hidden="true"
       :icon="icon"
-      size="small" />
+      :size="iconSize" />
```
spiess-demos, line 14: I'd suggest to also enable the second icon (that after the text) to be dynamically sized. I guess it is ok if we do not support two icon sizes within one button.

src/components/DpButton/DpButton.stories.mdx
```diff
@@ -23,7 +24,11 @@ import DpButton from './DpButton'
     variant: {
       options: ['solid', 'outline', 'subtle'],
       control: { type: 'select' }
-    }
+    },
+    iconSize: {
+          options: Object.keys(ICON_SIZES),
+          control: { type: 'select' }
+        }
```
spiess-demos, line 31:
```suggestion
    iconSize: {
      options: Object.keys(ICON_SIZES),
      control: { type: 'select' }
    }
```

src/components/DpButton/DpButton.vue
```diff
@@ -88,6 +89,16 @@ export default {
       default: ''
     },
 
+    /**
+     * Icon that will be placed before button text.
```
spiess-demos, line 93: See my previous comment (sizing for both icons).

```suggestion
     * Define the size of the button icons.
```

src/components/DpButton/DpButton.vue
salisdemos: Sure. good idea. 37965ebb33a7b063733091163747fc1859e74a1a

salisdemos:
> Nice! Apart from a few nitpicks, the changelog entry is missing.

1631db6bf1d2b25a3914e6ad04842f0a703e2f9c

src/components/DpButton/DpButton.vue
```diff
@@ -11,14 +11,15 @@
...
+      :size="iconSize" />
     <span
       :class="{'hide-visually': hideText}"
       v-text="text" />
     <dp-icon
       v-if="iconAfter"
       aria-hidden="true"
       :icon="iconAfter"
+      :size="iconSize"
       size="small" />
```
spiess-demos, line 23:
```suggestion
      :size="iconSize" />
```

salisdemos: if one do quick :-/

spiess-demos (approved):
Lgtm!

PR #371, opened 2023-07-13
fix: use correct translation values

spiess-demos:
We broke the test by changing the value of the placeholder option. Now we use the translation file to not have to duplicate strings here.


ahmad-demos (approved):
lgtm, codeReview, Thanks.

PR #372, opened 2023-07-13
Add notification icons to DpIcon

spiess-demos:
In an attempt do make DpIcon more resilient if called with a non-existent icon name, a few icons were added which we also need for the new design of the notifications.


salisdemos (approved):
Thanks. That would had helped me in the last week a lot. hopefully it will in the future too

PR #373, opened 2023-07-14
F tryout vue3 migration

salisdemos:
At some point, this will be an integrationbranch -- maybe -- or trash


spiess-demos (commented):
It seems like there are many code cleanup changes mangled into this PR, which are not necessarily tied to the Vue3 migration. To mitigate the risk of high "maintenance overhead" with this PR, we might separate all the changes that also do not break the current Vue2 setup into a cleanup PR that can be merged faster.

salisdemos:
close in favor of https://github.com/demos-europe/demosplan-ui/pull/575

PR #380, opened 2023-07-19
chore (refs T33878): Add DpTextArea to Storybook.

ahmad-demos:
https://yaits.demos-deutschland.de/T33878

- Add DpTextArea to Storybook.


src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+            label: 'Insert the Text.',
+            growToParent: true
+        }}>
+        {Template.bind({})}
+    </Story>
+</Canvas>
+
+### dataDpValidateIf
+
+Use to conditionally validate a required textarea field.
```
hwiem, line 66: I think we should explain here what kind of values can be passed to `dataDpValidateIf` so the usage becomes clear.

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+            label: 'Insert the Text.',
+            dataDpValidateIf: true
+        }}>
+        {Template.bind({})}
+    </Story>
+</Canvas>
+
+### reducedHeight
+
+Display the form field with reduced height (60px).
```
hwiem, line 82:
```suggestion
Display the textarea with reduced height (60px).
```

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+            label: 'Insert the Text.',
+            reducedHeight: true
+        }}>
+        {Template.bind({})}
+    </Story>
+</Canvas>
+
+### Required
+
+If a TextArea Field is `required`, an asterisk will be added to its label.
```
hwiem, line 98:
```suggestion
If a textarea is `required`, an asterisk will be added to its label.
```

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+### dataDpValidateErrorFieldname
+
+<Canvas>
+    <Story
+        name="dataDpValidateErrorFieldname"
+        args={{
+            id: 'textAreaId',
+            label: 'Insert the Text.',
+            dataDpValidateErrorFieldname: 'Text Message here'
```
hwiem, line 120:
```suggestion
            dataDpValidateErrorFieldname: 'Error message here'
```

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+        {Template.bind({})}
+    </Story>
+</Canvas>
+
+### Attributes
+
+Allows passing of additional attributes into the inner form element. Only chosen elements are allowed for each element type, failing if at least one attribute is not allowed within the array.
+Attributes are specified like :attributes="['attrName1=attrValue1', 'attrName2=attrValue2']".
+Params:
+element: String – Type of element
```
hwiem, line 131:
```suggestion
Allows passing of additional attributes into the inner form element (i.e., the textarea element). Allowed attributes for DpTextArea are
- rows
- cols

If at least one disallowed attribute is passed, an error will be thrown.
Only chosen elements are allowed for each element type, failing if at least one attribute is not allowed within the array.
Attributes are specified as follows:
```

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+For Examples:
```
hwiem, line 133: I'd remove this line

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+`:attributes="['rows=5', 'cols=10']"`
+
+<Canvas>
+    <Story
+        name="Attributes"
+        args={{
+            id: 'textAreaId',
+            label: 'Insert the Text.'
```
hwiem, line 142: Shouldn't we pass the `attributes` prop here?

hwiem (changes requested):
I don't know if this should be updated to Storybook 7 and then merged into the integration branch, but I think we could also do that in a second step.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Added
+- ([#380](https://github.com/demos-europe/demosplan-ui/pull/380)) Add documentation for DpTextArea ([@ahmad-demos](https://github.com/ahmad-demos))
```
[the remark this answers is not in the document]
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/380/commits/87c7841a90ab10acf95d6c56e286ce6d16353e5f

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,147 @@
...
+});
+
+### Disabled
+
+<Canvas>
+    <Story
+        name="Disabled"
+        args={{
+            id: 'textAreaId',
+            label: 'Insert the Text.',
```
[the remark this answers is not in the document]
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/380/commits/38ce76b1e88b5d32866319affcbdb62de8676686

src/components/DpTextArea/DpTextArea.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/380/commits/a5dfd404e9508f80f685043444495e3b8ea7d751

src/components/DpTextArea/DpTextArea.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/380/commits/16adc6f860bcd97023bc5bfbd489ae70750486b3

src/components/DpTextArea/DpTextArea.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/380/commits/ff5e2b7c917e1088c8e3ae991ca71cca46e10afc

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+
+- ([#380](https://github.com/demos-europe/demosplan-ui/pull/380)) Migrate DpTextArea stories to v7 ([@ahmad-demos](https://github.com/ahmad-demos))
```
salisdemos, line 9:
```suggestion
- ([#380](https://github.com/demos-europe/demosplan-ui/pull/380)) Add DpTextArea stories ([@ahmad-demos](https://github.com/ahmad-demos))
```

salisdemos (changes requested):

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,153 @@
...
+    </Story>
+</Canvas>
+
+### dataDpValidateIf
+
+Use to conditionally validate a required textarea field.
+
+This Prop take values `Boolean` or `String`.
+
+`data-dp-validate-if="input[name='uploadedFiles[videoSrc]']!=='', #videoTitle!==''"`
```
spiess-demos, line 70:
```suggestion
A condition consists of an expression that can be passed to `querySelector` (which should target another form element), followed by one of the operators `!==` or `===`, followed by a string to compare against. The operator and value can be omitted if the target element is of type checkbox or radio - in this case the textarea is validated if the referenced radio or checkbox element is checked.

Multiple conditions are separated by a comma.

In this example, a textarea is validated if a checkbox with the id `#send_to_address` is checked:
`data-dp-validate-if="#send_to_address"`

In this example, the textarea is validated if the input with the name `uploadedFiles[videoSrc]` is already filled, but the field with the id `videoTitle` is empty.
`data-dp-validate-if="input[name='uploadedFiles[videoSrc]']!=='', #videoTitle!==''"`
```

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,153 @@
...
+    <Story
+        name="dataDpValidateErrorFieldname"
+        args={{
+            id: 'textAreaId',
+            label: 'Enter some text',
+            dataDpValidateErrorFieldname: 'Error message here'
+        }}>
+        {Template.bind({})}
+    </Story>
+</Canvas>
```
spiess-demos, line 128:
```suggestion
### dataDpValidateErrorFieldname

Instead of displaying a custom error per field, as achieved by the custom data attribute `data-dp-validate-error` (which is not yet supported in DpTextArea), multiple failing fields may trigger one combined error notification that will then list the field names of each failing field. It reads such as `Bitte korrigieren Sie Ihre Eingabe in folgenden Feldern: Anrede, Betreff, Nachricht. Erst dann können Sie Ihre Angaben speichern.`. If not explicitly set, it defaults to the `label` prop, or, if not set, to null.

<Canvas>
    <Story
        name="dataDpValidateErrorFieldname"
        args={{
            id: 'textAreaId',
            label: 'Your Notes',
            dataDpValidateErrorFieldname: 'Notes'
        }}>
        {Template.bind({})}
    </Story>
</Canvas>
```

spiess-demos (approved):
When skimming over the docs, i realized that we need to improve two areas both for documentation and component implementation:

1. Some form components expose the `label` prop as object, featuring 'bold', 'hint', 'text', 'tooltip' properties inside; other components, like DpTextArea, have distinct props for those things... let alone DpEditor which does not use DpLabel at all. This is not just irritating when working with form components, but it prevents us from documenting `DpLabel` and its usage inside other form components once, and dropping the documnetation about label-related props from each component that uses it, by just cross-linking the `DpLabel` documentation. The implementation of the label component props could then also live inside `src/shared/props/label.js` to be shared by all form components that use `DpLabel` (which should be all of them).
2. The same can be said about form validation:
- it is not implemented coherently across form components
- it is not documented in Storybook
We could have a central definition of all validation-related props within `src/shared/props/dpValidate.js` which would be imported in every form component. We could then add a well-crafted documentation page for validation (at least the mixin), alongside the props that it is supporting, and cross-link from each form component to that instead of documenting each prop in each form component.

While this describes the target case, it must not be solved within this PR, so i'd say lgtm, lets merge it and do the above stuff later!

hwiem:
> When skimming over the docs, i realized that we need to improve two areas both for documentation and component implementation:
> 
> 1. Some form components expose the `label` prop as object, featuring 'bold', 'hint', 'text', 'tooltip' properties inside; other components, like DpTextArea, have distinct props for those things... let alone DpEditor which does not use DpLabel at all. This is not just irritating when working with form components, but it prevents us from documenting `DpLabel` and its usage inside other form components once, and dropping the documnetation about label-related props from each component that uses it, by just cross-linking the `DpLabel` documentation. The implementation of the label component props could then also live inside `src/shared/props/label.js` to be shared by all form components that use `DpLabel` (which should be all of them).
> 2. The same can be said about form validation:
> 
> * it is not implemented coherently across form components
> * it is not documented in Storybook
>   We could have a central definition of all validation-related props within `src/shared/props/dpValidate.js` which would be imported in every form component. We could then add a well-crafted documentation page for validation (at least the mixin), alongside the props that it is supporting, and cross-link from each form component to that instead of documenting each prop in each form component.
> 
> While this describes the target case, it must not be solved within this PR, so i'd say lgtm, lets merge it and do the above stuff later!

1. I agree completely.

2. also sounds reasonable, but I'm not sure if all form components use the same validation props. If they do, I agree; if they don't, it might make things a little more complicated, but I guess we could import only the props needed, so you know which ones to look up in the validation documentation, or we just list the props in the component documentation and link to the validation documentation.

src/components/DpTextArea/DpTextArea.stories.mdx
```diff
@@ -0,0 +1,161 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs'
+import DpTextArea from './DpTextArea'
```
spiess-demos, line 2: We can change this to 

```suggestion
import DpTextArea from './'
```

(see #515)

src/components/DpTextArea/DpTextArea.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import type { Meta, StoryObj } from '@storybook/vue'
+import DpTextArea from './DpTextArea.vue'
```
spiess-demos, line 2:
```suggestion
import DpTextArea from './'
```

spiess-demos (changes requested):

spiess-demos (approved):
Lgtm! We'll treat the points [mentioned here](https://github.com/demos-europe/demosplan-ui/pull/380#pullrequestreview-1595724044) later.

salisdemos (approved):

hwiem (approved):
lgtm

PR #381, opened 2023-07-20
Tweak z-index var names + values

spiess-demos:
- adapt utility classes to Tailwind naming
- use -10/+10 for z-indexes around zero to acommodate for the fact that there will be cases where users may squeeze something in there
- shorten scss var name according to the respective Tailwind naming


salisdemos (approved):
looks legit. thanks

PR #383, opened 2023-07-21
feat: Add new  DpTooltip and refactor v-tooltip Directive

salisdemos:
Since v-tooltip and VPopover is deprecated long time ago, we have to replace both of them.

With this PR, the v-tooltip-directive gets replaced and should work like before 
alongside we now have a DpTooltip which should replace VPopover. To get there VPopover is now marked as deprecated in our Code and after the new Component is available in demosplan-core, we can replace VPopover with DpTooltip.
IF that is done, we ca remove VPopover.


src/components/DpTootip/DpTooltip.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <component
+    :is="nodeType"
+    class="inline-block">
+    <slot />
```
spiess-demos, line 5: The `VPopover` component uses the default slot for the tooltip trigger, too, but it also exposes another slot named "popover" which receives the tooltip content.

src/components/DpTootip/utils/tooltip.js
```diff
@@ -0,0 +1,100 @@
+import { arrow, computePosition, flip, offset, shift } from '@floating-ui/dom'
```
spiess-demos, line 1: This implementation adds all needed markup right on mounted, and leaves the nodes in the dom until destroyed. This may lead to perf issues with lots of tooltip instances within one view, eg. in large lists with tooltips in it.

In the `VTooltip` implementation, the dom nodes representing the toolips were added whenever a tooltip was opened, and then, they were removed from the dom ca. 30 seconds after being hidden. I guess this was to ensure that when somebody hovers/unhovers a tooltip multiple times, the dom manipulations would not stack up.

spiess-demos (changes requested):

src/components/DpTootip/DpTooltip.vue
salisdemos: true. But we don't use it. thats why I thought we can keek it simple. I also did not enable passing all the fancy options. In my opinion we can enhance it, when we need it.

src/components/DpTootip/utils/tooltip.js
salisdemos: good point.... I will rewrite that

src/components/DpTootip/DpTooltip.vue
spiess-demos: > But we don't use it.

I found several occurrences of the slot with a search for `<template v-slot:popover>`.

src/components/DpTootip/utils/tooltip.js
salisdemos: I tried to find a better solution here 94d01bcb47d5a69010ef6b0ff0fe85eefa86e94e

src/components/DpTootip/DpTooltip.stories.mdx
```diff
@@ -0,0 +1,116 @@
...
+    }
+})
+
+
+# Tooltip
+
+Renders a Tooltip.
+
+
+## Usage without HTML
```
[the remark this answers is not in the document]
salisdemos: like so? ba1621c852c531382971aa61269da4ab071faa1b

src/components/DpTootip/DpTooltip.vue
salisdemos: its within here 94d01bcb47d5a69010ef6b0ff0fe85eefa86e94e
I could't test it by now because I have a problem with dplan and don't know how to use slots in storybook propperly

src/components/DpTootip/DpTooltip.stories.mdx
```diff
@@ -0,0 +1,116 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs'
+import { DpButton } from '../index'
+import DpTooltip from './DpTooltip.vue'
```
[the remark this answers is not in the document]
salisdemos: no ba1621c852c531382971aa61269da4ab071faa1b

salisdemos:
I hope I addressed all Your concerns...

src/components/DpTootip/DpTooltip.stories.mdx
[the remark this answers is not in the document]
spiess-demos: Yes. Also, we can tweak the docs after the storybook 7 PR is merged.

package.json
```diff
@@ -51,13 +52,12 @@
...
     "@typescript-eslint/eslint-plugin": "^5.4.0",
     "@typescript-eslint/parser": "^5.4.0",
-    "typescript": "~4.5.5",
     "@vue/cli-plugin-typescript": "~5.0.0",
     "@vue/eslint-config-typescript": "^9.1.0",
     "@vue/test-utils": "^1.3.5",
     "@vue/vue2-jest": "^29.2.4",
     "@webpack-cli/generators": "^3.0.0",
-    "autoprefixer": "^10.4.13",
+    "autoprefixer": "^10.4.14",
```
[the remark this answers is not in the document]
spiess-demos: > But I thought it doesn't hurt

We'll see.

src/components/DpTootip/utils/tooltip.js
```diff
@@ -0,0 +1,100 @@
...
+  Object.assign(tooltipEl.style, {
+    left: `${x}px`,
+    top: `${y}px`
+  })
+
+  /*
+   * to handle the position of the arrow.
+   * e.g. if the tooltip is on the top, we want to place the arrow at the bottom....
+   * `placement` can be 'bottom-start' etc s well, so we have to make sure to only take the first part
+   */
```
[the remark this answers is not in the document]
spiess-demos: @salisdemos what about this one?

src/components/DpTootip/utils/tooltip.js
spiess-demos: This looks good, tooltip elements are now created onHover instead of onMounted. However the `handleTimeoutForDestroy` thing does not what it is supposed to be, the element is not removed from dom.

salisdemos: no? In my tests it did remove them. Do You get any errors? (Its not removed instantly but after about 3  seconds after move out.)

src/components/DpTooltip/utils/tooltip.js
```diff
@@ -0,0 +1,121 @@
...
+  Object.assign(tooltipEl.style, {
+    left: `${x}px`,
+    top: `${y}px`
+  })
+
+  /*
+   * to handle the position of the arrow.
+   * e.g. if the tooltip is on the top, we want to place the arrow at the bottom....
+   * `placement` can be 'bottom-start' etc s well, so we have to make sure to only take the first part
+   */
```
salisdemos, line 100:
```suggestion
 /*
   * Handles the position of the arrow -  e.g. if the Tooltip is on the top, 
   * we want to place the arrow at the bottom, and so on. `placement` can be 
   * 'bottom-start' etc as well, so we have to make sure to only take the first part.
   */
```

src/components/DpTootip/utils/tooltip.js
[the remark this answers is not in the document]
salisdemos: For some reason it was presented me as outdated, so I thought it was already done. 
Changed in  16bd6b7e037935b2d2007a444ef4d77181978e18

src/components/DpTootip/utils/tooltip.js
spiess-demos: i'm testing it from within storybook.

src/components/DpTootip/DpTooltip.vue
spiess-demos: > I could't test it by now because I have a problem with dplan and don't know how to use slots in storybook propperly

Neither can i. Let's check this when demosplan can be accessed as usual, again!

spiess-demos (changes requested):
The directive seems to not work in the current implementation. I do not see the tooltip element added to the dom.

salisdemos:
got it. fixing one broke the other. Now it should work all over the place

spiess-demos (approved):
Lgtm & Works!

PR #379, opened 2023-07-19
Build(deps-dev): Storybook v7

salisdemos:
The setup is done. Now we "just" have to move the data-structure from v6 to v7. 
As example its done with the DpContextualHelp Component. 
... Just ~ 40 Files left

PR #396, opened 2023-07-27
F upgrade storybook migrate dp autocomplete

ahmad-demos:

- Migrate DpAutocomplete stories to v7.
- remove Translator.trans from DpAutocomplete

### DIFF
--- a/src/components/DpAutocomplete/DpAutocomplete.stories.mdx
+++ b/src/components/DpAutocomplete/DpAutocomplete.stories.mdx
@@ -2,7 +2,7 @@ import { Meta } from '@storybook/addon-docs'
 import DpAutocomplete from './DpAutocomplete'
 
 <Meta
-  title="Components/DpAutocomplete"
+  title="Components/Autocomplete"
   component={DpAutocomplete} />
 
 # Autocomplete
--- a/src/components/DpAutocomplete/DpAutocomplete.stories.mdx
+++ b/src/components/DpAutocomplete/DpAutocomplete.stories.mdx
@@ -2,7 +2,7 @@ import { Meta } from '@storybook/addon-docs'
 import DpAutocomplete from './DpAutocomplete'
 
 <Meta
-  title="Components/Autocomplete"
+  title="Components/DpAutocomplete"
   component={DpAutocomplete} />
 
 # Autocomplete
--- a/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
+++ b/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
@@ -17,6 +17,7 @@ export const Default: Story = {
     args: {
         height: '28px',
         label: 'label',
+        placeholder: 'Search',
         routeGenerator: function () {},
     }
 }
--- a/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
+++ b/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
@@ -4,7 +4,7 @@ import DpAutocomplete from './DpAutocomplete.vue'
 
 const meta: Meta<typeof DpAutocomplete> = {
     component: DpAutocomplete,
-    title: "Components/DpAutocomplete",
+    title: "Components/Autocomplete",
     argTypes: {
         'searchChanged': { action: 'searchChanged' }
     }
--- a/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
+++ b/src/components/DpAutocomplete/DpAutocomplete.stories.tsx
@@ -18,6 +18,6 @@ export const Default: Story = {
         height: '28px',
         label: 'label',
         placeholder: 'Search',
-        routeGenerator: function () {},
+        routeGenerator: () => '',
     }
 }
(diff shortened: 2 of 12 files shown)


src/components/DpAutocomplete/DpAutocomplete.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export default meta
+type Story = StoryObj<typeof DpAutocomplete>
+
+export const Default: Story = {
+    args: {
+        height: '28px',
+        label: 'label',
+        routeGenerator: function () {},
```
salisdemos, line 21: this function should return a (empty) string
```suggestion
        routeGenerator: () => '',
```

src/components/DpAutocomplete/DpAutocomplete.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+    }
+}
+
+export default meta
+type Story = StoryObj<typeof DpAutocomplete>
+
+export const Default: Story = {
+    args: {
+        height: '28px',
+        label: 'label',
```
salisdemos, line 19: we don't need to pass the default values.

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -55,7 +56,7 @@ export default {
     placeholder: {
       type: String,
       required: false,
-      default: () => Translator.trans('search')
+      default: () => de.placeholder
```
salisdemos, line 59: If You do things like this here, I would remove the unnecesarry method from label -> default too...
```
    label: {
      type: String,
      required: false,
      default: 'label'
    },
```

src/components/DpAutocomplete/DpAutocomplete.stories.mdx
```diff
@@ -2,7 +2,7 @@ import { Meta } from '@storybook/addon-docs'
 import DpAutocomplete from './DpAutocomplete'
 
 <Meta
-  title="Components/Autocomplete"
+  title="Components/DpAutocomplete"
```
salisdemos, line 5: A general Question: Do we want to have the `Dp`-Prefix in the Title/Navigation or not?

salisdemos (approved):

src/components/DpAutocomplete/DpAutocomplete.vue
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/396/commits/3ec68451d9f289f5e26008050ffc3ee52df18642

src/components/DpAutocomplete/DpAutocomplete.stories.tsx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/396/commits/c6bc9c28795611cfd6215e05c3ef69b2406b3649

src/components/DpAutocomplete/DpAutocomplete.stories.mdx
ahmad-demos: Yes we need, see in https://demosdeutschland.slack.com/archives/C5BA8ABB2/p1690353485332819?thread_ts=1690293841.102939&cid=C5BA8ABB2

When we make without `DP`-Prefix name, then the Docs will be separate from the default!

salisdemos: We only need the same Title in both files I guess. It shouldn't have anything todo with the Dp if I am right.

ahmad-demos: Yes you are right
Done in https://github.com/demos-europe/demosplan-ui/pull/396/commits/c7bab32955945945592fc6d128526305432cd559
Thanks!

spiess-demos: thanks for changing that!

src/components/shared/translations.js
```diff
@@ -5,6 +5,7 @@ const de = {
   errorMandatoryFieldsIntro: 'Bitte korrigieren Sie Ihre Eingabe in folgenden Feldern: ',
   errorMandatoryFieldsOutro: '. Erst dann können Sie Ihre Angaben speichern.',
   errorMandatoryFieldsDefault: 'Bitte füllen Sie alle Pflichtfelder korrekt aus. Erst dann können Sie Ihre Angaben speichern.',
+  placeholder: "Suche"
```
spiess-demos, line 8: I'd suggest to specify both the key and the value. "Suche" is too vague, and so is "placeholder".

```suggestion
  placeholderAutoSuggest: "Suchbegriff..."
```

spiess-demos (changes requested):

src/components/shared/translations.js
ahmad-demos: Done also in https://github.com/demos-europe/demosplan-ui/pull/396/commits/c176fd60bb45475146988f0460f8d82c0a3cc27d

spiess-demos (approved):
lgtm!

PR #397, opened 2023-07-28
F upgrade storybook migrate dp button

ahmad-demos:

- Migrate DpButton stories to v7.


src/components/DpButton/DpButton.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+        color: 'secondary',
+        hideText: false,
+        icon: 'delete',
+        iconSize: 'medium',
+        rounded: true,
+        text: 'Delete',
+        type: 'submit',
+        variant: 'outline'
+    }
+}
```
salisdemos, line 25: shouldn't be the Default quite "default". Here You set several props to not so default values ;-)

salisdemos (changes requested):

src/components/DpButton/DpButton.stories.tsx
ahmad-demos: https://github.com/demos-europe/demosplan-ui/pull/397/commits/3b0f3d620777ef82c0b2777830607ffa6910e3d5

salisdemos: but why do You define them at all? can't You just omit the props (they are all not required)

ahmad-demos: sorry, yes we can
https://github.com/demos-europe/demosplan-ui/pull/397/commits/a0fcebfe307df57f1b0ad9c37a874cbb69d2d5d1

thanks

salisdemos (approved):

PR #408, opened 2023-08-02
F upgrade storybook migrate dp multiselect

ahmad-demos:

- Migrate DpMultiselect story to v7.
- Add Trans-Key.


src/components/DpMultiselect/DpMultiselect.vue
```diff
@@ -27,7 +27,7 @@
         value
       }"
       :data-cy="dataCy"
-      v-dp-validate-multiselect="required"
+      :dp-validate-multiselect="required"
```
hwiem, line 30: Why this change?

ahmad-demos: because of `[Vue warn]: Failed to resolve directive: dp-validate-multiselect`

hwiem: That's because the directive needs to be imported and registered in the component:

```
import { dpValidateMultiselectDirective } from '../../lib/validation'

export default {
  name: 'DpMultiselect',

  components: {
    VueMultiselect
  },

  directives: {
    dpValidateMultiselectDirective
  }
...
}
```

and then it has to be used like this:

`v-dp-validate-multiselect-directive="required"`

I'd also suggest to get rid of the `directive` part in the name, but it doesn't have to be in this PR.

ahmad-demos: Thanks
Done in https://github.com/demos-europe/demosplan-ui/pull/408/commits/b8ab829b433068602f6cca1e253873b96222970f
And https://github.com/demos-europe/demosplan-ui/pull/412

hwiem (approved):
lgtm, code review only

PR #409, opened 2023-08-02
Fix(T34111): permanently opened dropdown

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34111

**Description:** This PR fixes issue with the permanently opened dropdown menu. After adding Tailwind CSS and adjusting new classes, the v-show directive with the style `'display: none;'` does not win in the specificity hierarchy. As a result, it has no impact on the flyout element:
- add a `hidden` class from the Tailwind library when the showFlyout variable has a negative value.


gruenbergerdemos (approved):
lgtm

PR #420, opened 2023-08-04
refactor(refs T31187): remove Translator usages

hwiem:
Ticket: https://yaits.demos-deutschland.de/T31187

- move all keys used in demosplan-ui that were previously defined in demosplan-core to `translations.js`
- remove Translator usages

### DIFF
--- a/src/components/DpContextualHelp/DpContextualHelp.vue
+++ b/src/components/DpContextualHelp/DpContextualHelp.vue
@@ -27,9 +27,9 @@ export default {
     tooltip: Tooltip
   },
 
-  computed: {
-    ariaLabel () {
-      return de.contextualHelp
+  data () {
+    return {
+      ariaLabel: de.contextualHelp
     }
   }
 }
--- a/src/components/DpContextualHelp/DpContextualHelp.vue
+++ b/src/components/DpContextualHelp/DpContextualHelp.vue
@@ -1,11 +1,12 @@
 <template>
   <i
     class="fa fa-question-circle"
-    :aria-label="Translator.trans('contextual.help')"
+    :aria-label="ariaLabel"
     v-tooltip="text" />
 </template>
 
 <script>
+import { de } from '../shared/translations'
 import { Tooltip } from '../../directives'
 
 export default {
@@ -24,6 +25,12 @@ export default {
 
   directives: {
     tooltip: Tooltip
+  },
+
+  computed: {
+    ariaLabel () {
+      return de.contextualHelp
+    }
   }
 }
 </script>
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -170,7 +170,8 @@
 </template>
 
 <script>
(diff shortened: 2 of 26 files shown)


ahmad-demos (approved):
lgtm, only codeReview
Thanks for clean code & good rebuilding for Trans-Key!

src/components/core/DpUpload/utils/UppyTranslations.js
```diff
@@ -10,11 +12,9 @@ const de = () => {
         browse: '{browse}',
         maxUploadSize: convertSize('GB', window.dplan.settings.maxUploadSize)
       }),
-      failedToUpload: Translator.trans('form.button.upload.failed', {
-        file: '{file}'
-      }),
+      failedToUpload: german.upload.failed,
```
ahmad-demos, line 15: I have a question, how to add `file: '{file}'` param here?

hwiem: We don't need to add it, it's replaced by uppy internally.

src/components/DpContextualHelp/DpContextualHelp.vue
```diff
@@ -24,6 +25,12 @@ export default {
 
   directives: {
     tooltip: Tooltip
+  },
+
+  computed: {
+    ariaLabel () {
+      return de.contextualHelp
+    }
```
salisdemos, line 33: I think a `data` is sufficent enough here. That avoids unnecessary eventlistener as far as I understand vue.
(would affect several places)

src/components/DpDataTable/DpDataTable.vue
```diff
@@ -381,7 +382,7 @@ export default {
         headerSelectHint: Translator.trans('aria.select.all'),
         lockedForSelection: Translator.trans('item.lockedForSelection'),
         searchNoResults: (searchTerm) => Translator.trans('search.no.results', { searchterm: searchTerm }),
-        tableLoadingData: Translator.trans('loading.data'),
+        tableLoadingData: de.loadingData,
```
salisdemos, line 385: I am not sure if I like accessing the transkeys directly.
If we do so now, we have to find all transkeys the moment we want to internationalize the lib.
I suggest to use something like `trans('loadingData')`

and the trans-method knows about the translation-file and adds the lang-flag

with that I hope it will be easier to implement vue-i18n later since it would be easier to find (and may replace) `trans(.*?)`

salisdemos (changes requested):

src/components/DpContextualHelp/DpContextualHelp.vue
hwiem: Thanks, changed in b2fd42ac0892bcb2d81712bb0f1a578e2ca46ae3 (I simply did it for all demosplan-ui)

src/components/DpDataTable/DpDataTable.vue
hwiem: It's a good point, but I think it might be more work to create a trans() function now than to simply remove the `de.` part later. Searching for 'de.' should give you all the trans keys, so I don't think it will require a lot of effort.

PR #438, opened 2023-08-16
Fix(T34173): flickering text when scrolling down

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34173

**Description:** This PR fixes the problem of text flickering when scrolling down. The problem arises when the element's container has limited space at the top and attempts to jump to the bottom. However, in this scenario, transitioning to the bottom is unnecessary, the element should be fixed at the top of the container. An extra check has been added wherever the function `_bindBottom` is called: when the value of the `stickToDirection` is 'top' the funciton `_bindBottom` will not be performed.

- when creating a new sticky object, set `'top'` as the value for `stickToDirection` in the DpTreeList component.


elmasdemos (changes requested):
I tested it and it didnt work this way. may we could test it together?

elmasdemos (commented):
code looks legitim, but with any reason it doesnt work when I try the change

hwiem (approved):
lgtm + works!

elmasdemos (approved):
lgtm, code review only

PR #445, opened 2023-08-18
refactor(refs T31187): remove DpDashboardTaskCard

hwiem:
Ticket: https://yaits.demos-deutschland.de/T31187

DpDashboardTaskCard has been moved to demosplan-core (https://github.com/demos-europe/demosplan-core/pull/1871) as it is too specific to keep it in demosplan-ui, so now we can remove it.

Tasks:

- [x] add changelog entry
- [x] merge https://github.com/demos-europe/demosplan-core/pull/1871 
- [ ] merge this PR

PR #453, opened 2023-08-21
refactor(refs T31187): move changelog entry

hwiem:
The changelog entry was accidentally added to the most recent release, but the changes (https://github.com/demos-europe/demosplan-ui/pull/445) were not actually included in it


ahmad-demos (approved):
lgtm, codeReview!

graupnerdemos (approved):
uh. thx. good catch

PR #460, opened 2023-08-22
fix: Adjust Changelog after auto-merge

salisdemos:
After merging main into `f_dp_tooltip` after a while of doing nothing, the changelog got mixed up.


spiess-demos (changes requested):
Well spotted! However now the entry is listed within the "Removed" section instead of the "Changed" section.

spiess-demos (approved):
lgtm!

PR #462, opened 2023-08-22
Consolidate tooltip components

spiess-demos:
Both DpContextualHelp and DpTooltipIcon do the same.
However, as the team agreed to go with DpContextualHelp,
it is now prepared to cater for all cases DpTooltipIcon
is used in:

- defaults to the question mark icon, but supports other icons
- defaults to medium size, but supports other icon sizes
- uses DpIcon instead of fontawesome


hwiem (approved):
lgtm, code review only

PR #473, opened 2023-08-25
F t32983 move generic components to demosplan  UI

muellerdemos:
### Ticket
https://yaits.demos-deutschland.de/T32983

### Description
These are really generic components and hence we should handle these components like other generic ones too, and add them to demosplan-ui
- There is an addon that needs to load an addon. Because I want to avoid copying the same files into multiple repositories, this solution made the most sense to me.

### linked PRs:
https://github.com/demos-europe/demosplan-core/pull/1486
https://github.com/demos-europe/demosplan-addon-demospipes/pull/79
https://github.com/demos-europe/demosplan-addon-segment-oracle/pull/1


gruenbergerdemos (commented):
`DpInsertableRecommendation` was moved [here](https://github.com/demos-europe/demosplan-core/pull/875) from demosplan-ui to demosplan, because it is not very generic and contains business logic. I'm not sure if we should move it back. Does it make sense to separate it from `DpRecommendationModal`which is still in core?

src/components/DpAddonWrapper/DpAddonWrapper.vue
```diff
@@ -0,0 +1,87 @@
+<template>
```
spiess-demos, line 1: I see the need to move DpAddonWrapper.vue into demosplan-ui... moving it to [demosplan-js-addon](https://github.com/demos-europe/demosplan-js-addon) would be more appropriate but as that repo does not provide a frontend build as of now i understand the reasoning.

However, as we plan to use demosplan-ui in a wider context, i feel that we must communicate the use case here, ecactly. Thats why i'd suggest to have a DpAddonWrapper.stories.mdx which explains the use case and gives a html example for how to use the component. I know that this is all documented within demosplan-core addon_documentation_frontend.md... maybe all we need is cross-linking to that documentation?

src/components/DpAddonWrapper/DpAddonWrapper.vue
```diff
@@ -0,0 +1,87 @@
...
+      dpRpc('addons.assets.load', { hookName: this.hookName })
+        .then(response => checkResponse(response))
+        .then(response => {
+          const result = response[0].result
+
+          for (const key of Object.keys(result)) {
+            const addon = result[key]
+            if (addon === undefined) {
+              /*
+               * If for some reason we don't receive a valid info object from the backend
```
spiess-demos, line 60: What is an "info object"?

src/components/DpAddonWrapper/DpAddonWrapper.vue
```diff
@@ -0,0 +1,87 @@
...
+              console.debug('Skipping addon hook response evaluation for ' + key)
+              continue
+            }
+
+            const contentKey = addon.entry + '.umd.js'
+            const content = addon.content[contentKey]
+
+            /*
+             * While eval generally is a BADIDEA, we really need to evaluate the code we're
+             * adding dynamically to use the provided addon's script henceforth.
```
spiess-demos, line 72:
```suggestion
             * While eval is generally a BAD IDEA, we really need to evaluate the code
             * we're adding dynamically to use the provided addon's script from now on.
```

src/components/DpBadge/DpBadge.vue
```diff
@@ -0,0 +1,57 @@
+<template>
```
spiess-demos, line 1: Please bootstrap minimal documentation:

- DpBadge.stories.mdx
- DpBadge.stories.tsx

Several simple examples can be found in existing components. I'll be happy to assist with further documentation content.

src/components/DpInsertableRecommendation/DpInsertableRecommendation.vue
```diff
@@ -0,0 +1,151 @@
+<template>
```
spiess-demos, line 1: Well... Using this component across two repositories does not make it a generic component at all. If we further go this way we'll make demosplan-ui the discount bin of frontend components, which is quite the opposite of what it should be in terms of usability and focus. 

In my opinion it would be ok to duplicate that component, if we need it in two places.

spiess-demos (changes requested):

src/components/DpAddonWrapper/DpAddonWrapper.vue
salisdemos: Can somebody explain the need of moving this file? I don't get the reason behind it.

muellerdemos: Well the currently developed addon needs to be implemented in the recommendation modal tabs but the recommendation modal as is is also an addon. Hence the addon needs to load an addon which is not a big issue from technically point of view but we need the wrapper to be accessible in the addon too. Hence I thought having the wrapper as part of demosplan-ui would solve this issue.

src/components/DpInsertableRecommendation/DpInsertableRecommendation.vue
muellerdemos: Allright, done here: https://github.com/demos-europe/demosplan-ui/pull/473/commits/8bfaee7960c1cdc19020cafd3470dc7e94e376c8

src/components/DpAddonWrapper/DpAddonWrapper.vue
muellerdemos: it should be ` a valid response..` sorry I just copy pasted this component before. Done here: https://github.com/demos-europe/demosplan-ui/pull/473/commits/88306d827a131b1bd75c029c4286b79dd5ddfbff

src/components/DpAddonWrapper/DpAddonWrapper.vue
salisdemos: But from an architektural point of view: do we want that? It sound quite bug-prone to me.

muellerdemos: For refs, we decided it would be better to refactor the recommendation modal in its corresponding addon to be a tab too and provide a generic modal in the core which loads both addons.

src/components/DpBadge/DpBadge.vue
muellerdemos: We decided it should be better if the recommendation modal will be refactored to a generic modal which is able to load both addons. This way the addon warpper can remain in the core. https://github.com/demos-europe/demosplan-ui/pull/473/commits/9e235a911cbe304196bd256283c3f63d45a85c3f but with the changes you mentioned in this PR

spiess-demos: But DpBadge can very well live in demosplan-ui!

muellerdemos: Sorry my bad I was just refering to the addon wrapper- Added the bootstrapped stories for the DpBadge component here:
https://github.com/demos-europe/demosplan-ui/pull/473/commits/d93a3fa6216d242bc6dd5950d59500959e7667ee

salisdemos (approved):
legit

PR #475, opened 2023-08-28
feat(refs T27593): add tailwind to storybook

spiess-demos:
to further enable component development within storybook, tailwind styles are made available in storybook.

Most tutorials on this topic refer to the deprecated @storybook/addon-styling, however @storybook/addon-styling-webpack installs way less dependencies and seems to do the same job.

As we are now able to build tailwind in demosplan-core, we can remove the safelist hack that forced tw to render all css classes by default.


salisdemos (approved):
lovely

PR #476, opened 2023-08-28
fix: Clear Bugs after merging all together

salisdemos:
I had to fix several Bugs to get tiptap running without throwing errors.

One Major issue:
we have to expose prosemirror stuff in demosplan-ui to prevent initializing multiple instances with different versions of PM when using PM direct without tiptap


spiess-demos (commented):
I think that this is not the best way to handle versioning conflicts of prosemirror.

A first thought: demosplan-ui is bundled, so why is there a conflict at all?

A second thought: making all prosemirror modules that are shared by demosplan-ui and demosplan-core peer depencencies of demosplan-ui should then do the trick as well (hopefully).

spiess-demos (approved):
lgtm! (code review only)

PR #478, opened 2023-08-28
feat(refs T27593): use correct webpack config for tw

spiess-demos:
the new api of addon-styling-webpack is different from the old one, so postcss must be configured more explicitly. See https://github.com/storybookjs/addon-styling-webpack/issues/1


ahmad-demos:
now storybook is working but default page is empty!

ahmad-demos:
> now storybook is working but default page is empty!

after `yarn install` working

ahmad-demos (approved):

elmasdemos (approved):
lgtm code review

PR #491, opened 2023-08-30
chore: add DpStickyElement to Storybook.

ahmad-demos:
- add DpStickyElement to Storybook.


src/components/DpStickyElement/DpStickyElement.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+    title="Components/StickyElement"
+    component={DpStickyElement}
+/>
+
+# StickyElement
+
+The StickyElement component to add static position any container, menu or slide in the (top or bottom)
+of Page when scrolling.
+
+## Props:
```
spiess-demos, line 14: Some general thoughts on the documentation of component props, and which content should go here (in the *.stories.mdx page):

As we're already documenting the props via DocBlocks in the actual component code, and these DocBlocks are rendered in the controls panel of the actual story page, i think it is not needed to list any prop here, again. We should rather focus on improving existing DocBlocks and filling in those which are not already existing. That being said i'd opt to remove props documentation in the *.stories.mdx pages completely.

Instead, we may provide a high level description on what the component does, and in which cases to use it. The Docs page may further feature a basic example for those who are not willing to dive into the actual stories.

spiess-demos (changes requested):

src/components/DpStickyElement/DpStickyElement.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/491/commits/b2964dc76dda8a6816229b387d45615c4ee4cadd

src/components/DpStickyElement/DpStickyElement.stories.mdx
```diff
@@ -0,0 +1,32 @@
...
+      :current="currentPage"
+      :total="totalPages"
+      @page-change="handlePageChange" />
+    <dp-select-page-item-count
+      @changed-count="setPageItemCount"
+      :page-count-options="itemsPerPageOptions"
+      :current-item-count="itemsPerPage"
+      :translations="{ pagerElementsPerPage: Translator.trans('pager.per.page') }" />
+  </div>
+</dp-sticky-element>
```
spiess-demos, line 31: For the code examples in general, i'd suggest to only use the minimal code that is needed to get this component working. In this case, the `ref` and `class` attributes are not needed for the example, as they may distract readers.

If a code example needs additional components to illustrate usage, i suggest to also keep that example markup to a bare minimum: only use native Html elements, or components from demosplan-ui. Also i feel that those "placeholder components" should be reduced to their bare minimum. In the example here, we pass a div into the slot, which contains two rather complex components. In my opinion, it would be enough to have two button elements (maybe "Save" and "Cancel"), to illustrate the case of a sticky save bar. It would also be ok to omit the example content at all, by only placing a Html comment inside like `<!-- Whatever you place here is gonna stick to the bottom -->`.

spiess-demos (changes requested):

src/components/DpStickyElement/DpStickyElement.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/491/commits/947a8cc7bf1ff2b7c4b02134759a4d2f496973a9
Thanks!

PR #498, opened 2023-08-31
fix: fix storybook errors, update version.

ahmad-demos:
- fix storybook errors, update version.

for test:
`yarn install` ->` yarn storybook`

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "postcss-loader": "^7.0.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "storybook": "7.0.27",
+    "storybook": "7.4.0",
     "storybook-addon-vue-slots": "^0.9.21",
     "style-dictionary": "^3.7.1",
     "style-loader": "^3.3.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -96,7 +96,7 @@
     "postcss-loader": "^7.0.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
-    "storybook": "7.0.27",
+    "storybook": "7.4.0",
```
spiess-demos, line 99: This fixes the bug, thanks for spotting it!

However, as Storybook is not a dependency (for production code) it can also be moved into the devDependencies block.

package.json
```diff
@@ -159,5 +159,8 @@
     "build:prod": "webpack --mode=production --define-process-env-node-env=production",
     "test": "jest",
     "watch": "webpack --watch"
+  },
+  "resolutions": {
```
spiess-demos, line 163: This change can be reverted with the above fix. I just tested it, there is no error even without the `resolution` block.

spiess-demos (changes requested):

package.json
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/498/commits/23fcc761bc85dcb7077673d1a288a7e603cec03f

package.json
spiess-demos: sorry for confusing you, i actually overlooked that this is actually the devDependencies block

spiess-demos (approved):
lgtm!

PR #506, opened 2023-09-04
chore: Replace axios with native fetch api

salisdemos:
When moving from axios to fetch, we had to keep the api as it was before....

This is an alternative aproach to https://github.com/demos-europe/demosplan-ui/pull/549
Because we prefer to get rid of axios in favor of using the native fetch api, I close [#549 ](https://github.com/demos-europe/demosplan-ui/pull/549)


src/components/DpButton/DpButton.vue
```diff
@@ -42,7 +42,7 @@ export default {
 
   props: {
     /**
-     * When waiting for an axios response, the visual state of the button can be changed via the `busy` property.
+     * When waiting for an response, the visual state of the button can be changed via the `busy` property.
```
spiess-demos, line 45:
```suggestion
     * While waiting for a network response, the visual state of the button can be changed via the `busy` property.
```

src/lib/DpApi.js
```diff
@@ -123,8 +129,8 @@ const handleResponseMessages = function (responseMeta) {
  * Perform rudimentary response validation, handle response messages.
  *
  * @param {Object} response
- *                    The axios wrapper around the XMLHttpRequest instance.
- *                    See https://github.com/axios/axios#response-schema for documentation.
+ *                    The Response Provided by the fetch API around the XMLHttpRequest instance.
```
spiess-demos, line 132:
```suggestion
 *                    The response provided by the fetch API around the XMLHttpRequest instance.
```

spiess-demos (commented):
just grammar nitpicks.

spiess-demos (approved):
lgtm (code review only).

src/lib/DpApi.js
```diff
@@ -27,38 +26,46 @@ const api2defaultHeaders = {
...
+
+  if (method.toUpperCase() !== 'GET') {
+    payload.body = JSON.stringify(data)
+  } else if (options.serialize === true) {
+    delete payload.options.serialize
+
+    url = `${url}?${stringify(params, { encodeValuesOnly: true, arrayFormat: 'brackets' })}`
+  }
+
+  const response = await fetch(url, payload)
```
hwiem, line 55: Shouldn't we throw an error in case the fetch fails?

src/lib/DpApi.js
```diff
@@ -78,29 +85,28 @@ const dpRpc = function (method, parameters, id = null) {
...
 }
 
 /**
  * Perform an external API call without any default headers
  */
-const externalApi = function (url) {
-  const contentType = axios.defaults.headers.common['Content-Type']
-  delete axios.defaults.headers.common['Content-Type']
+const externalApi = async (url, data) => {
+  const response = await fetch(url, data)
```
hwiem, line 107: Shouldn't we wrap this in a try/catch block to be able to handle an error, should it occur?

src/lib/DpApi.js
```diff
@@ -123,8 +129,8 @@ const handleResponseMessages = function (responseMeta) {
  * Perform rudimentary response validation, handle response messages.
  *
  * @param {Object} response
- *                    The axios wrapper around the XMLHttpRequest instance.
- *                    See https://github.com/axios/axios#response-schema for documentation.
+ *                    The response provided by the fetch API around the XMLHttpRequest instance.
```
hwiem, line 132: Should the "around the XMLHttpRequest instance." part still be here? I don't understand the meaning :thinking:

package.json
```diff
@@ -36,7 +36,6 @@
     "@uppy/progress-bar": "^3.0.0",
     "@uppy/tus": "^3.0.1",
     "a11y-datepicker": " ^0.9.0",
-    "axios": "^0.27.2",
```
hwiem, line 39: Shoudn't there be an updated `yarn.lock`, too?

graupnerdemos (changes requested):
+1 on everything @hwiem already pointed out.

package.json
salisdemos: true 14152f743723a784ecdd18b67422106eb25e1969

src/lib/DpApi.js
salisdemos: I don't know why I added the `.json`. I removed it again. With that change imho we don't need a try catch here anymore, since in case of an error the method will return it and it can be handled in the calling method.

src/lib/DpApi.js
salisdemos: cood catch b6fc5709ec131e3a6e3cad7fac3c1a9ebf23664d

src/lib/DpApi.js
salisdemos: a1b124f472872fd5f170944b77f98418ae0d5f6b

CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ### Changed
 - ([#537](https://github.com/demos-europe/demosplan-ui/pull/537)) Rename DpNotifyMessage to DpNotification ([@ahmad-demos](https://github.com/@ahmad-demos))
 - ([#374](https://github.com/demos-europe/demosplan-ui/pull/374)) Bump Tiptap from v1 to v2 ([@salisdemos](https://github.com/salisdemos))
+- ([#506](https://github.com/demos-europe/demosplan-ui/pull/506)) Replace internal from DpApi/DpRpc. Move from axios to native fetch ([@salisdemos](https://github.com/salisdemos))
```
hwiem, line 52: This needs to be moved to the "unreleased" section now

hwiem (changes requested):

CHANGELOG.md
salisdemos: thanks for spotting this
a51a322dbaff56dbd7d6b6c8c3dd1f28588fce6e

hwiem (approved):
Lgtm, code review only

graupnerdemos (approved):

PR #511, opened 2023-09-05
fix(refs T34566): Pass propper Values to init Tooltips

salisdemos:
**Ticket** https://yaits.demos-deutschland.de/T34566

When using the tooltip-directive It was only supportet to pass the content. 
But we also have situations, where we pass Objects with container, content or classes. 
That was forgotten when rewriting the tooltip directive. 
So we have to reimplement it now


spiess-demos (approved):
lgtm & works!

PR #515, opened 2023-09-07
Use aliases in components

spiess-demos:
With https://github.com/demos-europe/demosplan-ui/pull/513, we introduced the internal alias `~` (which points to `./src`), along with an index.ts in every component folder. This PR adds index files for all components and uses the `~` where appropriate.

Note that the error `TypeError: The provided value is not of type 'Element'.` within tests does not originate from this PR, it is present within current main branch, too.


ahmad-demos (approved):
lgtm, only codeReview!

ahmad-demos (approved):

PR #521, opened 2023-09-12
chore: add DpDatepicker to Storybook.

ahmad-demos:
- add DpDatepicker to Storybook.

### DIFF
--- a/src/components/DpDatepicker/DpDatepicker.stories.mdx
+++ b/src/components/DpDatepicker/DpDatepicker.stories.mdx
@@ -8,11 +8,14 @@ import DpDatepicker from './'
 
 # Datepicker
 
-The Datepicker component use `a11y-datepicker` component.
+The Datepicker component allows users to enter a date either by typing or by selecting a date from the calendar.
+
+Use a date picker when you want to provide a view that allows the user to select a calendar date.
 
 ## Default usage
 
 ```html
 <dp-datepicker
-  id="datepicker_id" />
+  id="datepicker_id"
+  v-model="planDate" />
 ```
--- a/src/components/DpDatepicker/DpDatepicker.stories.mdx
+++ b/src/components/DpDatepicker/DpDatepicker.stories.mdx
@@ -1,0 +1,18 @@
+import { Meta } from '@storybook/addon-docs'
+import DpDatepicker from './'
+
+<Meta
+    title="Components/Datepicker"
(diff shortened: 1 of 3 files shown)


src/components/DpDatepicker/DpDatepicker.stories.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+# Datepicker
+
+The Datepicker component use `a11y-datepicker` component.
+
+## Default usage
+
+```html
+<dp-datepicker
+  id="datepicker_id" />
```
spiess-demos, line 17: To have a more realistic example, i'd suggest to add `v-model="planDate"` here.

src/components/DpDatepicker/DpDatepicker.stories.mdx
```diff
@@ -0,0 +1,18 @@
...
+import DpDatepicker from './'
+
+<Meta
+    title="Components/Datepicker"
+    component={DpDatepicker}
+/>
+
+# Datepicker
+
+The Datepicker component use `a11y-datepicker` component.
```
spiess-demos, line 11: The first paragraph of the documentation of each component should express these things:

- what the component does (aka. what the component enables users to do)
- when to use the component (a helpful question here would be: for which occasions do we use the component)

A good example can be found in the DpDetails documentation.

I think we do not have to mention that `a11y-datepicker` is used internally, as this is an implementation detail that users of `DpDatepicker` do not have to care about.

spiess-demos (changes requested):

src/components/DpDatepicker/DpDatepicker.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/521/commits/1c48425c019ec7a0de619a4b3ccf39b2a36a0925

src/components/DpDatepicker/DpDatepicker.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/521/commits/1c48425c019ec7a0de619a4b3ccf39b2a36a0925

spiess-demos (approved):
Lgtm, thanks... that's what i meant, basically!

PR #523, opened 2023-09-13
Fix(T34571): place tooltip in the foreground

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34571

**Description:** There are cases when tooltips are in the modal windows and have a smaller z-index than the modal.

- added a method that checks the z-index of the parent element and appends `zIndex + 1` to the style of the tooltip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,11 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Changed
-- ([#523](https://github.com/demos-europe/demosplan-ui/pull/523)) Place the tooltip in the foreground using the 'z-ultimate' class in the DpContextualHelp.vue component ([@sakutademos](https://github.com/sakutademos))
 - ([#374](https://github.com/demos-europe/demosplan-ui/pull/374)) Bump Tiptap from v1 to v2 ([@salisdemos](https://github.com/salisdemos))
 
 ### Added
 
+- ([#523](https://github.com/demos-europe/demosplan-ui/pull/523)) Added a method that checks the z-index of the parent element and appends zIndex + 1 to the style of the tooltip ([@sakutademos](https://github.com/sakutademos))
 - ([#521](https://github.com/demos-europe/demosplan-ui/pull/521)), ([#522](https://github.com/demos-europe/demosplan-ui/pull/522)), ([#524](https://github.com/demos-europe/demosplan-ui/pull/524)) Add DpDatepicker, DpDatetimePicker & DpDateRangePicker documentation to Storybook ([@ahmad-demos](https://github.com/@ahmad-demos))
 - ([#518](https://github.com/demos-europe/demosplan-ui/pull/518)) Add DpUpload documentation to Storybook ([@ahmad-demos](https://github.com/@ahmad-demos))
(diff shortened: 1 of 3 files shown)


spiess-demos (changes requested):
This implementation would stack the tooltip always on top of everything else. This may be the desired behavior most of the time, but in cases where elements with higher hierarchy partly overlap a tooltip, the trigger element would then be stacked below the overlapping element (eg. a sidebar) but the tooltip would appear above.

I suggest to use a solution [based on this implementation](https://github.com/demos-europe/demosplan-core/blob/22e6a44496bd491cab280d8578bc83b3f33a164d/client/js/components/map/map/DpOlMapEditFeature.vue#L210-L217) but instead of setting a utility class we could just set a style attribute for the z-index that equals the z-index of the trigger element plus 1. This way we ensure that tooltips are always displayed above their respective trigger elements but below other elements of higher hierarchy.
As this implementation needs access to the trigger element, it needs to be implemented within the `initTooltip` and `showTooltip` methods of the Tooltip directive.

sakutademos:
> This implementation would stack the tooltip always on top of everything else. This may be the desired behavior most of the time, but in cases where elements with higher hierarchy partly overlap a tooltip, the trigger element would then be stacked below the overlapping element (eg. a sidebar) but the tooltip would appear above.
> 
> I suggest to use a solution [based on this implementation](https://github.com/demos-europe/demosplan-core/blob/22e6a44496bd491cab280d8578bc83b3f33a164d/client/js/components/map/map/DpOlMapEditFeature.vue#L210-L217) but instead of setting a utility class we could just set a style attribute for the z-index that equals the z-index of the trigger element plus 1. This way we ensure that tooltips are always displayed above their respective trigger elements but below other elements of higher hierarchy. As this implementation needs access to the trigger element, it needs to be implemented within the `initTooltip` and `showTooltip` methods of the Tooltip directive.

Refactored it [here]( https://github.com/demos-europe/demosplan-ui/pull/523/commits/8f170bc1f696020c5827eda9fa144a18c4ed7a36). Did you mean this?

spiess-demos (approved):
Lgtm (code review only).

Non-blocking nitpick: one could argue if this is to be listed within "Added" (as per the added convenience of not having to manually define different z-index classes anymore) or within "Changed" (because of the changed behavior of the Tooltip directive).

spiess-demos:
> > This implementation would stack the tooltip always on top of everything else. This may be the desired behavior most of the time, but in cases where elements with higher hierarchy partly overlap a tooltip, the trigger element would then be stacked below the overlapping element (eg. a sidebar) but the tooltip would appear above.
> > I suggest to use a solution [based on this implementation](https://github.com/demos-europe/demosplan-core/blob/22e6a44496bd491cab280d8578bc83b3f33a164d/client/js/components/map/map/DpOlMapEditFeature.vue#L210-L217) but instead of setting a utility class we could just set a style attribute for the z-index that equals the z-index of the trigger element plus 1. This way we ensure that tooltips are always displayed above their respective trigger elements but below other elements of higher hierarchy. As this implementation needs access to the trigger element, it needs to be implemented within the `initTooltip` and `showTooltip` methods of the Tooltip directive.
> 
> Refactored it [here](https://github.com/demos-europe/demosplan-ui/pull/523/commits/8f170bc1f696020c5827eda9fa144a18c4ed7a36). Did you mean this?

Exactly. Thanks for taking this on!

PR #526, opened 2023-09-14
chore: add DpButtonIcon to Storybook.

ahmad-demos:
- add DpButtonIcon to Storybook.

### DIFF
--- a/src/components/DpButtonIcon/DpButtonIcon.vue
+++ b/src/components/DpButtonIcon/DpButtonIcon.vue
@@ -37,7 +37,7 @@ export default {
 
   computed: {
     translatedText () {
-      return Translator.trans(this.text)
+      return this.text
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/components/DpButtonIcon/DpButtonIcon.vue
```diff
@@ -37,7 +37,7 @@ export default {
 
   computed: {
     translatedText () {
-      return Translator.trans(this.text)
+      return this.text
```
spiess-demos, line 40: The computed only exists because of the Translator. You can as well use the prop in the template directly when dropping Translator.

spiess-demos (changes requested):
As the component is only used within one addon, and the DpButton component has evolved since to be capable of rendering icon-only buttons, this component should be deprecated. See https://github.com/demos-europe/demosplan-ui/blob/79f9d9b5f6eb96a36c9c8d90e110788df2aa6344/src/components/DpTooltipIcon/DpTooltipIcon.vue#L11

src/components/DpButtonIcon/DpButtonIcon.vue
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/526/commits/3dc2e1494d09cc79169c32d0d4582d3793567d8a

PR #530, opened 2023-09-15
Fix(T34571): place tooltip in the foreground 

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T34571

**Description:** There are cases when tooltips are in the modal windows and have a smaller z-index than the modal.

- added a method that checks the z-index of the parent element and appends zIndex + 1 to the style of the tooltip


spiess-demos (approved):
lgtm (code review)

PR #532, opened 2023-09-15
chore: add DpToggleForm to Storybook.

ahmad-demos:
- add DpToggleForm to Storybook.


spiess-demos (changes requested):
Can we move DpToggleForm to demosplan-core? I'd argue that this is not a generic component.

ahmad-demos:
> Can we move DpToggleForm to demosplan-core? I'd argue that this is not a generic component.

Yes, we can!

PR #539, opened 2023-09-21
Fix(T34716) tooltip update method

sakutademos:
**Description:** This PR adds a method that updates the tooltip value.

- it removes the previous events from the wrapper element
- adds the showTooltip function with a new value to the wrapper
- checks if the tooltip element already exists in the HTML (if yes: removes it from the HTML and destroys the tooltip)
- creates a new tooltip with a new value
- adds new events to the wrapper element

However, it does not fix the current bug because the tooltip can only be updated if it appears in the HTML and can be found using `getElementById`.


spiess-demos:
> However, it does not fix the current bug because the tooltip can only be updated if it appears in the HTML and can be found using `getElementById`.

Did you try to change the directive hook from ` inserted` to `mounted`?

PR #542, opened 2023-09-21
Fix(T34716): add an update method in the tooltip directory

sakutademos:
**Ticket:**  https://yaits.demos-deutschland.de/T34716

**Description:** This PR adds a method that updates the tooltip value.

### DIFF
--- a/src/components/DpTooltip/utils/tooltip.js
+++ b/src/components/DpTooltip/utils/tooltip.js
@@ -5,6 +5,7 @@ import { v4 as uuid } from 'uuid'
 let handleShowTooltip = null
 let handleHideTooltip = null
 let handleTimeoutForDestroy = null
+let tooltips = []
 
 const deleteTooltip = (tooltipEl) => {
   if (tooltipEl) {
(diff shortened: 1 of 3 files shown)


src/components/DpTooltip/utils/tooltip.js
```diff
@@ -5,9 +5,11 @@ import { v4 as uuid } from 'uuid'
 let handleShowTooltip = null
 let handleHideTooltip = null
 let handleTimeoutForDestroy = null
+let tooltips = []
 
 const deleteTooltip = (tooltipEl) => {
   if (tooltipEl) {
+    tooltips.filter(el => el.id !== tooltipEl.id)
```
salisdemos, line 12: what for this is needed?

src/components/DpTooltip/utils/tooltip.js
```diff
@@ -59,11 +63,14 @@ const initTooltip = (el, value, options) => {
 
   const id = `tooltip-${uuid()}`
   const zIndex = getZIndex(el)
+  const tooltipData = { id: id, value: value }
+
+  tooltips.push(tooltipData)
```
salisdemos, line 68: why do You use an array and not an Object?
```suggestion
  tooltips[id] = tooltipData
```
with that you can avoid the loop in line 145

salisdemos (changes requested):

src/components/DpTooltip/utils/tooltip.js
sakutademos: refactored it here https://github.com/demos-europe/demosplan-ui/pull/542/commits/f4102553846455c57f7e55a9f4174056425a5cf8. Use an object instead of an array

src/components/DpTooltip/utils/tooltip.js
sakutademos: refactored it https://github.com/demos-europe/demosplan-ui/pull/542/commits/f4102553846455c57f7e55a9f4174056425a5cf8

salisdemos (approved):
thanks. Looks good

PR #545, opened 2023-09-25
chore: Fix DpTab Storybook error.

ahmad-demos:
- Fix DpTab Storybook error.
`this.tabs[0] is undefined`

### DIFF
--- a/src/components/DpTabs/DpTabs.stories.tsx
+++ b/src/components/DpTabs/DpTabs.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/vue'
 import DpTabs from './'
-import DpTab from './'
+import DpTab from './DpTab.vue'
 
 const meta: Meta<typeof DpTabs> = {
     component: DpTabs,


salisdemos (approved):

src/components/DpTabs/DpTabs.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import type { Meta, StoryObj } from '@storybook/vue'
 import DpTabs from './'
-import DpTab from './'
+import DpTab from './DpTab.vue'
```
salisdemos, line 3: why do we need a explicit import for DpTab but not for DpTabs

salisdemos (changes requested):

src/components/DpTabs/DpTabs.stories.tsx
ahmad-demos: DpTabs imported from index.ts
```
export { default } from './DpTabs.vue'
export * from './DpTabs.vue'
```

salisdemos: ok. So we could do it for tab.vue the same way?

ahmad-demos: there can only ever be one default export!

ahmad-demos: @spiess-demos

spiess-demos: as we need both components within the same story, but the `import x from ./` syntax only works for default exports, which we cannot have two at the same time within one file, the solution `import DpTab from './DpTab.vue'` is ok imho.

salisdemos (approved):

PR #546, opened 2023-09-25
snapshot test for the DpContextualHelp component

sakutademos:
**Description:** Remove the snapshot test for the `DpContextualHelp` component due to its dynamically changing `aria-describedby` value, which currently causes failures. Since we require a demosplan-ui release, it could be skipped for now and fixed later.


salisdemos (approved):
Sad but true

PR #548, opened 2023-09-26
chore: WIP: Add DpTableCard to Storybook

ahmad-demos:
WIP: Add DpTableCard to Storybook

### DIFF
--- a/src/components/core/DpTableCardList/DpTableCard.stories.mdx
+++ b/src/components/core/DpTableCardList/DpTableCard.stories.mdx
@@ -1,0 +1,45 @@
+import { Meta } from '@storybook/addon-docs'
+import DpTableCard from './DpTableCard'
+
+<Meta
+    title="Components/TableCard"
+    component={DpTableCard}
+/>
+
+# TableCard
+
+The TableCard component is used to display tabular data in a visually organised and structured way.
+
+It's often used to present data in a list, making it easier for users to scan and understand the information.
+
+## Custom usage
+
+```html
+<ul class="o-list o-list--card u-mb">
+  <div class="o-list__item">
+    <dp-table-card class="o-accordion u-ph-0_5" v-bind="args">
+      <template v-slot:header>
+        <div class="flex items-start">
+          <div class="relative z-above-zero u-mt-0_75">
+              <input
+                  type="checkbox">
+          </div>
+          <button
+              @click="args.isOpen = !args.isOpen"
+              type="button"
+              data-cy="userListItemToggle"
+              class="btn--blank o-link--default u-pv-0_75">
+            Click to Toggle
+          </button>
+        </div>
+      </template>
+      <div>
+        Test 1
+      </div>
+      <div>
+        Test 2
+      </div>
+    </dp-table-card>
+  </div>
+</ul>
+```
--- a/src/components/core/DpTableCardList/DpTableCard.stories.tsx
+++ b/src/components/core/DpTableCardList/DpTableCard.stories.tsx
@@ -1,0 +1,68 @@
+import type { Meta, StoryObj } from '@storybook/vue'
+import DpTableCard from './DpTableCard.vue'
+
+interface IDpTableCard {
+    open: boolean
+    isOpen: boolean
+}
+
+const meta: Meta<typeof DpTableCard> = {
+    component: DpTableCard,
+    title: 'Components/TableCard',
+    render: (args) => ({
+        components: {
+            DpTableCard
+        },
+        setup() {
+            return { args }
+        },
+        template: `
+          <div>
+            <ul class="o-list o-list--card u-mb">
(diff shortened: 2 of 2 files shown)


spiess-demos:
I got concerns regarding the existence of the DpTableCard component in demosplan-ui. Do we really want to keep it as a distinct component? We have DpDataTable to display tabular data (which we could add a "dark" theme to, if we want to support it). We have DpAccordion for collapsible content in general, and we have DpCard (which does add to the confusion in terms of naming). DpTableCard is used within the user bundle and (in a light theme) within the draftStatement lists (where it does not work well as user tests suggest). 

If we want to keep it around, which would be ok given that we decide to further spread its usage, i'D suggest to rename it to "DpCardListItem" and make it a child of "DpCardList" - which also removes the dangerously hidden semantic need of the parent element being an `<ul>`.

salisdemos:
> If we want to keep it around, which would be ok given that we decide to further spread its usage, i'D suggest to rename it to "DpCardListItem" and make it a child of "DpCardList" - which also removes the dangerously hidden semantic need of the parent element being an `<ul>`.

I absolutely agree, that we have to consolidate Components  here. I am not sure how, but three for more or less the same thing sounds too much.

can we somehow find out, what kind of things we want to display in some kind of card and where the differences have to be? 
Right now all three Components do different things, but I assume that's more historical than a wanted curcumstance.

src/components/core/DpTableCardList/DpTableCard.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+The TableCard component is used to display tabular data in a visually organised and structured way.
+
+It's often used to present data in a list, making it easier for users to scan and understand the information.
+
+## Custom usage
+
+```html
+<div>
+    <ul class="o-list o-list--card u-mb">
+      <div class="o-list__item">
```
salisdemos, line 20: the component itself resolves in a li. so the parent should be a ul.
Do we need the div here?

src/components/core/DpTableCardList/DpTableCard.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        components: {
+            DpTableCard
+        },
+        setup() {
+            return { args }
+        },
+        template: `
+          <div>
+            <ul class="o-list o-list--card u-mb">
+              <div class="o-list__item">
```
salisdemos, line 21: see bove

salisdemos (changes requested):

src/components/core/DpTableCardList/DpTableCard.stories.mdx
ahmad-demos: thanks
Done in https://github.com/demos-europe/demosplan-ui/pull/548/commits/0a8415e5b3023fe483afc217ae0e6ab54cfaf3af

src/components/core/DpTableCardList/DpTableCard.stories.tsx
ahmad-demos: thanks
Done in https://github.com/demos-europe/demosplan-ui/pull/548/commits/0a8415e5b3023fe483afc217ae0e6ab54cfaf3af

spiess-demos:
> > If we want to keep it around, which would be ok given that we decide to further spread its usage, i'D suggest to rename it to "DpCardListItem" and make it a child of "DpCardList" - which also removes the dangerously hidden semantic need of the parent element being an `<ul>`.
> 
> I absolutely agree, that we have to consolidate Components here. I am not sure how, but three for more or less the same thing sounds too much.

We should consolidate components, but not necessarily here. We had the plan to [move some curated components to demosplan-ui](https://yaits.demos-deutschland.de/T27599), but then we moved everything over in a hurry. I would argue that if in doubt (that this is a component we want within demosplan-ui), we should move it over to demosplan-core again. The more components we have here, the more we have to maintain.

> can we somehow find out, what kind of things we want to display in some kind of card and where the differences have to be? Right now all three Components do different things, but I assume that's more historical than a wanted curcumstance.

Sure we would have to do that, but i do not want it to be a blocker for demosplan-ui 1.0. So until we have the time to refine DpCardList, i suggest to move it back into core.

salisdemos:
is this still WIP like the title and Description says?

ahmad-demos:
> is this still WIP like the title and Description says?

Now PR is ready!

src/components/core/DpTableCardList/DpTableCard.stories.tsx
salisdemos: this was not the DIV i meant.
I was talking about `<div class="o-list__item">`

spiess-demos (changes requested):
As i said above: i opt to move the whole component to demosplan-core.

PR #558, opened 2023-10-05
fix: Adjust Test after Refactoring

salisdemos:
and add missing dependency
Where ever that comes from

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@demos-europe/demosplan-ui",
-  "version": "0.1.14",
+  "version": "0.1.16",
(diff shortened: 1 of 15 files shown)


gruenbergerdemos (approved):
lgtm

package.json
```diff
@@ -98,6 +98,7 @@
     "react-dom": "^18.2.0",
     "storybook": "7.4.0",
     "storybook-addon-vue-slots": "^0.9.21",
+    "string-width": "^4.2.3",
```
spiess-demos, line 101: Why is this dependency added here? We do not use it anywhere.

PR #572, opened 2023-10-11
fix: rebuild yarn.lock

spiess-demos:
As recommended in https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092 rebuilding yarn.lock seems to fix the bug described in the ticket. As a side effect, some patch updates were applied.


salisdemos (approved):
:heart:

PR #575, opened 2023-10-11
chore(refs T19759): WIP Make demosplan-ui work for demosplan-core vue3 Compat mode

salisdemos:
Adjust everything neccessary to make demosplan-core work in vue3 compat mode. 
This will be just a part of the movement to vue3 in demosplan-ui.

see https://github.com/demos-europe/demosplan-core/pull/1151


salisdemos:
This has no diff to main.

PR #585, opened 2023-10-12
fix(refs T34978): adjust Method Path

salisdemos:
To make it work that data gets pasted into the current text, the used method was undefined. So we had to adjust the path to the method which changed when migrating to tiptap 2

along I shortened some imports path to address hints from sonar cloud

### DIFF
--- a/src/components/core/DpEditor/DpEditor.vue
+++ b/src/components/core/DpEditor/DpEditor.vue
@@ -330,7 +330,7 @@ import {
   Text,
   Underline
 } from './libs/tiptapExtensions'
-import { CleanHtml, Tooltip } from '../../../directives'
+import { CleanHtml, Tooltip } from '~/directives'
 import {
   Editor, // Wrapper for prosemirror state
   EditorContent, // Renderless content element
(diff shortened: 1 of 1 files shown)


src/components/core/DpEditor/DpEditor.vue
```diff
@@ -477,16 +477,16 @@ export default {
         const suggestionGroupSchema = {
           matcher: {
             char: '@|$|#', // A single char that should trigger a suggestion
-            allowSpaces: true || false,
-            startOfLine: true || false
+            allowSpaces: 'boolean',
+            startOfLine: 'boolean'
```
spiess-demos, line 481: Here and lines 488-489: why are you changing this?

salisdemos: damn'd. This was just a tryout because. SonarCloud mentioned something. It was not supposed to be in the PR. I revert it...
544581f3e9a979d7852f74d3f32d3627c8253168

spiess-demos (approved):
lgtm & works!

PR #589, opened 2023-10-16
chore (refs T34964): Add DpTextWrapper to Storybook.

ahmad-demos:
https://yaits.demos-deutschland.de/T34966

- Add DpTextWrapper to Storybook.


spiess-demos (changes requested):
The DpTextWrapper component was introduced to render Vue components that are referenced inside dynamically loaded Html, as with DpObscure and other components that are part of statement- or fragment text. Whenever those are loaded (or reloaded) via api the components should mount again.

See [Commit](https://github.com/demos-europe/demosplan/commit/f733d95ed327d3458728ca2da7d7b81acdb5c85c), [Ticket](https://yaits.demos-deutschland.de/T8564).

I'd argue that both DpTextWrapper and DpHeightLimit should be moved to demosplan-core as those are highly specific to the way we used to work with statatement texts in the past. Those components are not used in recent components anyways.

PR #591, opened 2023-10-18
refactor(refs T29004): re-add label tests

hwiem:
Ticket: https://yaits.demos-deutschland.de/T29004

- the label tests (that were disabled after refactoring the label props to be a single object) are fixed and enabled again
- some more tests that were disabled in DpInput.spec.js for unknown reasons are enabled again as well

### How to test

Run the tests, there should be no error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+- ([#591](https://github.com/demos-europe/demosplan-ui/pull/591)) Add label tests for form components ([@hwiem](https://github.com/hwiem))
+
 ## v0.2.1 - 2023-10-18
 
 ### Changed
--- a/tests/form/DpCheckbox.spec.js
+++ b/tests/form/DpCheckbox.spec.js
@@ -10,6 +10,8 @@ describe('DpCheckbox', () => {
     }
   })
 
+  runLabelTests(wrapper)
+
   const checkbox = wrapper.find('input[type="checkbox"]')
   runBooleanAttrTests(wrapper, checkbox, 'required')
   runBooleanAttrTests(wrapper, checkbox, 'disabled')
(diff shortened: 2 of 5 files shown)


salisdemos (approved):
Thank You!

spiess-demos (commented):
Why do we have to run the label tests in all other component tests, too? isn't it sufficient to test the label component for itself?

hwiem:
> Why do we have to run the label tests in all other component tests, too? isn't it sufficient to test the label component for itself?

That's a very good question. By running them in other components we ensure that the props are passed properly from the outer component to DpLabel, but maybe that's not really the purpose of unit tests?

spiess-demos:
> By running them in other components we ensure that the props are passed properly from the outer component to DpLabel, but maybe that's not really the purpose of unit tests?

As `runLabelTests` sets the props by itself, this is not exactly true. Testing child components in a component test looks (at least a little bit) like testing the implementation. To ensure that each form component has a working label implementation we could change `runLabelTests` so that it takes the props in question (`text`, `hint`, `required` and so on) as a second param, and then expexts the rendered html to contain those values within their expected html nodes.

hwiem:
> > By running them in other components we ensure that the props are passed properly from the outer component to DpLabel, but maybe that's not really the purpose of unit tests?
> 
> As `runLabelTests` sets the props by itself, this is not exactly true. Testing child components in a component test looks (at least a little bit) like testing the implementation. To ensure that each form component has a working label implementation we could change `runLabelTests` so that it takes the props in question (`text`, `hint`, `required` and so on) as a second param, and then expexts the rendered html to contain those values within their expected html nodes.

I'm not sure I understand your concern. Why is it better to pass the props as a parameter?

spiess-demos:
> I'm not sure I understand your concern. Why is it better to pass the props as a parameter?

Thanks for clarifying the `runLabelTests` approach in an 1:1. I simply did not look close enough, the test already does what i thought it should do.

spiess-demos (approved):
lgtm

PR #592, opened 2023-10-18
feat: group tiptap and storybook dependabot PRs

spiess-demos:
All tiptap (and storybook) dependencies will have to be updated at once, otherwise things will break. So here we go.


salisdemos (approved):
Awesome

PR #605, opened 2023-10-20
F t34965 add dp editable list to storybook

ahmad-demos:
https://yaits.demos-deutschland.de/T34965

- Add Trans-Keys to DpEditableList.
- Add DpEditableList to Storybook.

Screenshot 2023-10-23 at 14-48-04 Components _ EditableList - Default ⋅ Storybook
Screenshot 2023-10-23 at 14-48-14 Components _ EditableList - Default ⋅ Storybook


src/components/shared/translations.js
```diff
@@ -85,6 +85,14 @@ const de = {
   obscure: {
     title: 'Dieser Text wurde als geschwärzt markiert, um Datenschutzrichtlinien zu entsprechen.'
   },
+  operations: {
```
salisdemos, line 88: I think its a good Idea to open a section "operations". But then we should move `save` and `select` into here too.

salisdemos: and we already have `cancel: "Abbrechen"`. We shouldn't double this with `abort: "Abbrechen"`.

salisdemos (changes requested):

src/components/shared/translations.js
ahmad-demos: Thanks
Done in [chore (refs T34965): Add Trans-Keys to DpEditableList.](https://github.com/demos-europe/demosplan-ui/pull/605/commits/f2d17a472341e6446fbfeacacbc798d36019ec77)

salisdemos (approved):

PR #606, opened 2023-10-20
fix: lock dependency to omit string-width error

spiess-demos:
A mysterious hoisting behavior leads to the ERR_REQUIRE_ESM error from the transitive dependency "string-width" with `yarn test` or `yarn storybook`. A fix suggests to lock "jackspeak" to 2.1.1.

See https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092


spiess-demos:
> > the simplest solution here is to upgrade to Yarn 3.Could that be a solution for us?
> 
> Could that be a solution for us?

i tried it (as a bugfix some weeks ago) and ran into severe issues. this is not a quick solution i guess.

PR #613, opened 2023-10-24
chore: Deprecate Usage of translator in data-dp-validate-error

salisdemos:
There should be no need to translate here. It should be passed translated already


hwiem (commented):
Can't we just fix the usages of Translator with data-dp-validate-error and assume it will not appear again?

salisdemos:
> Can't we just fix the usages of Translator with data-dp-validate-error and assume it will not appear again?

Maybe. But that would have been more work ;-P
Maybe we don't merge this one and I try to force the changes in the core-repository.

PR #615, opened 2023-10-25
F t34968 add dp bulk edit header to storybook

ahmad-demos:
https://yaits.demos-deutschland.de/T34968
https://github.com/demos-europe/demosplan-core/pull/2245

- WIP: Add DpBulkEditHeader to Storybook.
- Add Tran-Key to DpBulkEditHeader component.

Screenshot 2023-11-01 at 13-04-28 Components _ BulkEditHeader - Default ⋅ Storybook
Screenshot 2023-11-01 at 13-04-39 Components _ BulkEditHeader - Default ⋅ Storybook


src/components/core/DpBulkEditHeader.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+    component={DpBulkEditHeader}
+/>
+
+# BulkEditHeader
+
+The BulkEditHeader component allows users to perform bulk editing operations on a set of data or records.
+
+In a data management system, users often need to edit multiple records simultaneously.
+Instead of editing each record individually, a bulk edit functionality allows users to select multiple records
+and apply changes to specific fields in one go.
```
spiess-demos, line 15: These paragraphs can be shortened, and complemented with an explanation of the features of the component:

```suggestion
The BulkEditHeader component allows users to perform bulk editing operations on a set of records. It is typically used in conjunction with `DpDataTable` or a similar component that allows users to select multiple records. Then, actions to be performed on that selection can be placed inside the default slot of `DpBulkEditHeader`.

The component features a visually prominent style and showcases the number of selected items, along with a button to reset the current selection.
```

src/components/core/DpBulkEditHeader.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+      selection-text="Elements"
+      @reset-selection="resetSelection"
+      v-bind="args">
+    <button
+        class="btn-icns u-m-0"
+        type="button"
+        @click.prevent="deleteSelection">
+      <i class="fa fa-times u-mr-0_125" />
+      Delete
+    </button>
```
spiess-demos, line 33: - Can you use `DpButton` in the example?
- The `fa-times` icon should not be used for destructive actions (see the Storybook pawge for DpIcon), please use the `icon` prop of `DpButton` with the correct icon

src/components/core/DpBulkEditHeader.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+  </dp-bulk-edit-header>
+  <input
+      type="checkbox"
+      @click="toggleSelectAll"> Select all
```
spiess-demos, line 37: Even if, or better, just because this is an example, we should deploy accessibility best practices here and use the `label` element.

src/components/core/DpBulkEditHeader.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+  <div
+      class="u-mt-0_5"
+      v-for="item in args.options"
+      :key="item.id">
+    <input
+        :id="item.id"
+        v-model="item.checked"
+        :checked="!item.checked === false"
+        type="checkbox"/>
+    {{ item.label }}
```
spiess-demos, line 47: see other comment (accessibility/label).

src/components/core/DpBulkEditHeader.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+            }
+        },
+        setup() {
+            const toggleSelectAll = () => {
+                for (let element of args.options) {
+                    if(element.checked === false) {
+                        element.checked = true
+                    } else {
+                        element.checked = false
+                    }
```
spiess-demos, line 23: This logic inverts the checked state for each element, instead of just setting checked to true for each element.

src/components/core/DpBulkEditHeader.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    element.checked = false
+                }
+            }
+
+            const deleteSelection = () => {
+                for (let element of args.options) {
+                    if(element.checked === true) {
+                        args.options = args.options.filter(checked => checked.checked != true)
+                    }
+                }
```
spiess-demos, line 38: As this is just an example, i would simply call `resetSelection()` here, instead of deleting the elements.

src/components/core/DpBulkEditHeader.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            return { args, deleteSelection, resetSelection, toggleSelectAll }
+        },
+        template: `
+          <div>
+              <dp-bulk-edit-header
+                  v-if="selectedItemsCount.length > 0"
+                  :selected-items-count="selectedItemsCount.length"
+                  selection-text="Elements"
```
spiess-demos, line 48: This is a weakness of the component that i'd like to fix, maybe within this same PR: The text that is actually displayed is [computed within DpBulkEditHeader](https://github.com/demos-europe/demosplan-ui/blob/041435553f49b4692b99696a68e58493066460f3/src/components/core/DpBulkEditHeader.vue#L42), from the two props that are also passed to DpBulkEditHeader. However this way we are not able to generate custom plural forms. I'd suggest to have just one prop `selectedItemsText` (and drop the `selectedItemsCount` in DpBulkEditHeader). Then we would move the count to the translations in demosplan-core, to enable nice plural forms.

src/components/core/DpBulkEditHeader.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+type Story = StoryObj<IDpBulkEditHeader>
+
+export default meta
+
+export const Default: Story = {
+    args: {
+        options: [
+            { label: 'Option 1', id: '1', checked: false },
+            { label: 'Option 2', id: '2', checked: false },
+            { label: 'Option 3', id: '3', checked: false }
```
spiess-demos, line 91:
```suggestion
            { label: 'Item 1', id: '1', checked: false },
            { label: 'Item 2', id: '2', checked: false },
            { label: 'Item 3', id: '3', checked: false }
```

src/components/core/DpBulkEditHeader.vue
```diff
@@ -38,6 +40,10 @@ export default {
   computed: {
     selectedItemsText () {
       return this.selectedItemsCount + ' ' + this.selectionText
+    },
+
+    unselect () {
+      return de.operations.deselect.element
```
spiess-demos, line 46: I'd call the computed `deselect` as well.

src/components/shared/translations.js
```diff
@@ -86,7 +86,8 @@ const de = {
     add: 'Hinzufügen',
     delete: 'Löschen',
     deselect: {
-      all: 'Alle abwählen'
+      all: 'Alle abwählen',
+      element: 'Auswahl aufheben'
```
spiess-demos, line 90: The terms are basically expressing the same thing. `operations.deselect.all` is used in DpMultiSelect and DpTreeList for the same thing. I would prefer the term "Auswahl aufheben" over "Alle abwählen", but i'd use the key `operations.deselect.all` and drop the other key.

spiess-demos (changes requested):

src/components/core/DpBulkEditHeader.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/627fc1d2bf10ff7a9610d1976a2a4c74cc3003d9

src/components/core/DpBulkEditHeader.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/627fc1d2bf10ff7a9610d1976a2a4c74cc3003d9

src/components/core/DpBulkEditHeader.stories.tsx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/e2d7e2eefabf8df90725a47df489e4c9a2a92f16

src/components/core/DpBulkEditHeader.vue
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/9a6ce2821e733a9d19fa5c3bf4e5b0e8eca82c70

src/components/core/DpBulkEditHeader.stories.mdx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/eab8db3c048044e96042470da14f70d37f88c82b

src/components/shared/translations.js
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/e768e6fbc47ad63a2e2be07bbbc708b9be23bb45

src/components/core/DpBulkEditHeader.stories.tsx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/e241f51b2ce6408cc249b8fca407a82e0549c68b

src/components/core/DpBulkEditHeader.stories.tsx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/cb41626c68f765620949f45d396957f14379cd3b

src/components/core/DpBulkEditHeader.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+        template: `
+          <div>
+              <dp-bulk-edit-header
+                  v-if="selectedItemsCount.length > 0"
+                  :selected-items-text="selectedItemsCount.length + ' ' + 'Elements selected'"
+                  @reset-selection="resetSelection"
+                  v-bind="args">
+                <dp-button
+                    text="Delete"
+                    icon="fa-trash"
```
spiess-demos, line 44: There is no icon `fa-trash` in DpIcon. This is why the `warning` icon is rendered as a fallback. The `delete` icon is more appropriate here.

src/components/core/DpBulkEditHeader.stories.mdx
spiess-demos: - see comment in DpButton
- please also use DpButton here

spiess-demos (changes requested):

src/components/core/DpBulkEditHeader.stories.tsx
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/aff06c048674a339549873b0989a933239e33357

src/components/core/DpBulkEditHeader.stories.tsx
ahmad-demos: see also: https://github.com/demos-europe/demosplan-core/pull/2245

src/components/core/DpBulkEditHeader.stories.tsx
spiess-demos: Please look at the docs in `http://localhost:6006/?path=/docs/components-icon--docs#delete` (local storybook) how it is used.

ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/1aa637355f94cc7c8f030c88ec64db5319bacc71

src/components/core/DpBulkEditHeader.stories.mdx
```diff
@@ -0,0 +1,50 @@
...
+```html
+<div>
+  <dp-bulk-edit-header
+      v-if="selectedItemsCount.length > 0"
+      :selected-items-text="selectedItemsCount.length + ' ' + 'Elements selected'"
+      @reset-selection="resetSelection"
+      v-bind="args">
+    <dp-button
+        text="Delete"
+        icon="fa-delete"
```
spiess-demos, line 26:
```suggestion
        icon="delete"
```

spiess-demos (dismissed):

src/components/core/DpBulkEditHeader.vue
```diff
@@ -5,39 +5,32 @@
...
-      default: ''
+      required: true
     }
   },
 
   computed: {
-    selectedItemsText () {
-      return this.selectedItemsCount + ' ' + this.selectionText
+    deselect () {
+      return de.operations.deselect.all
```
salisdemos, line 33: this don't have to be a computed. It should be `data`. Or maybe a prop with a default value.

salisdemos (changes requested):
Just one little thing

src/components/core/DpBulkEditHeader.vue
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/615/commits/4624c5b63418c5d63cc6507befd195f69f52516f

salisdemos: Why did You not move it to `props` ? Did You thought about it?

spiess-demos: imho this **must not** be a prop, otherwise we would have to implement props for every translation key within demosplan-ui. i thought the plan is to solve that on a library wide level, maybe when adding vue-i18n.

salisdemos (approved):

PR #620, opened 2023-10-31
docs: be clear about xmark vs. delete icon

spiess-demos:


hwiem (commented):
Nice, thanks! I'm only wondering if we should add a changelog entry for this?

spiess-demos:
> Nice, thanks! I'm only wondering if we should add a changelog entry for this?

I think we should not, as this is rather a minor change.

hwiem (approved):

PR #621, opened 2023-10-31
chore: bump vue and vue-template-compiler to 2.7.15

spiess-demos:
As the peer dependency needs to be in sync with vue-template-compiler bc jest tests throw an error otherwise, and bc demosplan-core has bumped vue to 2.7.15 as well, we can bump both to 2.7.15 here, as well.

Closes #608.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,7 @@ updates:
       tiptap:
         patterns:
           - "@tiptap*"
+      vue:
+        patterns:
+          - "vue"
+          - "vue-template-compiler"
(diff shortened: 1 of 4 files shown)


salisdemos:
> As the peer dependency needs to be in sync with vue-template-compiler 

Then we should group then straight away.

salisdemos (changes requested):
> As the peer dependency needs to be in sync with vue-template-compiler 

Then we should group them straight away.

spiess-demos:
> > As the peer dependency needs to be in sync with vue-template-compiler
> 
> Then we should group them straight away.

I have also thought of that. However the `vue@^2.6` entry referenced in yarn.lock is a transitive dependency that (at least this is my assumtion) is not respected by dependabot. However lets see what happens: 79c35a45aa4cd9a783a0c14e69a78978704f502b.

salisdemos (approved):

PR #623, opened 2023-11-06
fix(refs T19759): Make demosplan-ui work in vue3 compat mode level 2

salisdemos:
make Code adjustments, to make the compat-mode run when used in demosplan-core.


src/lib/Stickier.js
```diff
@@ -36,8 +36,8 @@ class Stickier {
   ) {
     // Set up variables from options
     this.element = stickyElement
-    this.context = contextElement || this.element.offsetParent
-    this.container = containerElement || this.element.offsetParent
+    this.context = contextElement || this.element.parentElement
+    this.container = containerElement || this.element.parentElement
```
spiess-demos, line 40: Why did you chose to use [parentElement](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement) here? It always returns the parent element, regardless if it is positioned or not (contrary to [offsetParent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) which returns the nearest positioned element up the DOM hierarchy). This change will probably break the positioning logic in `_stickier()`.

spiess-demos (changes requested):

src/lib/Stickier.js
salisdemos: because I had an issue with offsetParent, wich resulted to undefind. But thanks for the hint. then I have to investigate deeper...

src/components/DpDataTable/DpTableHeader.vue
```diff
@@ -30,7 +30,7 @@
         :header-field="hf"
         :idx="idx">
         <slot
-          v-if="$slots[`header-${hf.field}`][0].children.length > 0"
+          v-if="$slots[`header-${hf.field}`]()[0].children.length > 0"
```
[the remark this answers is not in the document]
salisdemos: with vue3 slots are now functions. don't ask we why
https://v3-migration.vuejs.org/breaking-changes/slots-unification.html

package.json
```diff
@@ -49,6 +49,7 @@
     "tippy.js": "^6.3.7",
     "uuid": "^9.0.0",
     "v-tooltip": "2.1.3",
+    "vue": "3.1.5",
```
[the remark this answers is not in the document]
salisdemos: because its a peer that need to be shipped by the leading project.
But, true, It doesn't have to be moved. I think in the End it shouldn't make any difference. I will role it back.

src/components/DpDataTable/DpTableHeader.vue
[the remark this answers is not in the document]
spiess-demos: :bulb:  thanks for the explanation.

package.json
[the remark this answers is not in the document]
salisdemos: f76ad9ab92522c740309aac4354fdb5ed6aa20fd

src/lib/Stickier.js
salisdemos: Thanks. Found the reason on the other end. Vue3 changes the behaviour of $el which leaded to a differnet element.
revered here 89002a1434aeccc9de2801af040cdd7d989ac497

webpack.config.js
```diff
@@ -41,6 +41,7 @@ const config = {
     'tippy.js',
     'uuid',
     'v-tooltip',
+    'vue',
```
spiess-demos, line 44: We do not need to set Vue as external here if we do not have it as a dependency.

spiess-demos (changes requested):

webpack.config.js
salisdemos: hmm.. I doublecheck that. But I needed it for some reason...

spiess-demos: > hmm.. I doublecheck that. But I needed it for some reason...

you moved vue into the `dependencies` block... maybe it was included in the build, then?

salisdemos: That was my thought too. But I want to be sure.

hwiem (approved):
lgtm

webpack.config.js
salisdemos: tested. Looks like You where right 9d449c0814abc8dd85d62a77a470a10f2fc1dbaa

salisdemos: ok. Got it. 
We need it. 
We don't need it to build demosplan-ui, but we need it to tell the using app, that we need vue.
We wouldn't need that if we would have it in the `dependencies` section. BUt as `devDependency` it get stripped out during the build and we have to set it as `external`. And thats what we want - imho.

spiess-demos: > Shouldn't it be a peerDependency?

thats what i think, too, overthinking it once more.

spiess-demos (approved):
lgtm, code review.

PR #624, opened 2023-11-06
fix: Use proper import for buildSuggestion

salisdemos:
What ever happend there, This was wrong. Now it works...


spiess-demos (approved):
lgtm! code review only

PR #629, opened 2023-11-07
fix(refs T35139): Allow dynamically use slots available in vue-multiselect

salisdemos:
If we have to / want to use a specific slot in the multiselect, it wasn't possible by now.
with this changes, we can define any slot that is available in vue-multiselect (https://vue-multiselect.js.org/#sub-slots) to use it.

To test this, You have to add a slot (like singleLabel) in some occurence, where You use dp-multiselect. The Content should be rendered.

**Ticket** https://yaits.demos-deutschland.de/T35139


package.json
```diff
@@ -98,7 +98,7 @@
     "react-dom": "^18.2.0",
     "storybook": "7.5.3",
     "storybook-addon-vue-slots": "^0.9.21",
-    "style-dictionary": "^3.7.1",
+    "style-dictionary": "3.8.0",
```
sakutademos, line 101: should it be part of this PR?

salisdemos: no. thanks. reverted them.

sakutademos (approved):
code review + works, lgtm

PR #630, opened 2023-11-08
bug(refs T34956): Fix stickier position

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T34956

If the element ends below context the position should be the bottom of context. Otherwise the element has not enough space and is not reachable. Don't stick to direction in this case. 

It seems like the bug was introduced with this fix https://github.com/demos-europe/demosplan-ui/pull/438
Since this doesn't fix the bug completely (see https://yaits.demos-deutschland.de/T34173) I suggest to revert it and to try to fix it another way.


spiess-demos (approved):
I'd follow your proposal to revert the other fix. @sakutademos would you agree here?

PR #633, opened 2023-11-08
Fix(T33788): disabled scrolling after closing the modal window

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T33788

**Description:** This PR addresses the issue where scrolling becomes impossible after closing the modal window:

- call the `preventScroll` function in the `toggle() `method for cases where the `@after-leave` transition does not work correctly, especially when there is a parent component (such as a specialized modal like AssignEntityModal, ConsolidateModal or CopyStatementModal)

- [x] Link all relevant tickets
- [x] Move the tickets on the board accordingly


spiess-demos (approved):
lgtm! however after dropping the assessment table we could change this back...

PR #638, opened 2023-11-09
Cleanup package.json

spiess-demos:
- remove @storybook/addon-mdx-gfm which was only needed when migrating from storybook 6
- remove special jackspeak resolution as we found the fix for the "string-width error" (running `yarn upgrade jest`)
- define engines for npm and node to keep demosplan-ui in sync with demosplan-core

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,6 @@
     "webpack-cli": "^5.0.0"
   },
   "engines": {
-    "npm": ">= 6.0.0",
     "node": ">= 16.18.1"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "webpack-cli": "^5.0.0"
   },
   "engines": {
-    "node": ">= 16.18.1"
+    "node": ">= 18.13.0"
   },
   "peerDependencies": {
     "prosemirror-history": "^1.3.0",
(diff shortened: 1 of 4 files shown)


salisdemos (approved):

package.json
```diff
@@ -111,6 +110,10 @@
     "webpack-bundle-analyzer": "^4.7.0",
     "webpack-cli": "^5.0.0"
   },
+  "engines": {
+    "npm": ">= 6.0.0",
```
graupnerdemos, line 114: we don't technically need to define npm here since we're not actually using npm.

package.json
```diff
@@ -111,6 +110,10 @@
...
+    "node": ">= 16.18.1"
```
graupnerdemos, line 115: shouldn't this be `>= 18.18.2`?

graupnerdemos (changes requested):

package.json
spiess-demos: well.. this is just copied from demosplan-core. however, 6b3e9e39f1c4c96a645b827a082ff3210004b20e

package.json
spiess-demos: Why do you say so?

spiess-demos: 6d03e46d12ee3d37ea6de52d86e2bc5aa46436ee i set it to the version used in containers

graupnerdemos (approved):
lgtm

PR #640, opened 2023-11-13
fix: Use proper method to reset content in DPEditor

salisdemos:
setContent is part of commands.


spiess-demos (approved):
lgtm! code review only.

PR #643, opened 2023-11-14
fix(refs T19759): fix warnings/errors [vue 3 upgrade]

hwiem:
**Integration PR:** #575
**Ticket:** https://yaits.demos-deutschland.de/T19759

Fixes some console warnings/errors that occur when accessing a dplan project in the browser.

- use new hooks in directives:
in vue 3, the hooks in directives have changed:
`bind` -> `beforeMount`
`inserted` -> `mounted`
`update` -> `updated`
`componentUpdated` -> `updated`
`unbind` -> `unmounted`
see https://v3-migration.vuejs.org/breaking-changes/custom-directives.html#_3-x-syntax

- pass null instead of false as attribute value
to remove html attributes, they need to be set to `null` instead of `false` in vue 3
see https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html

- use `slots` instead of `scopedSlots` (`scopedSlots` have been removed)

- remove usage of `$listeners` (in vue 3, listeners are added to `$attrs` and no longer available as `$listeners`)

- use deep watcher for array (in vue 3, watchers for arrays are not deep by default)

- `beforeDestroy` hook has been renamed to `beforeUnmount`

### Related PRs
- [x] #623 (Needs to be merged first)


src/components/DpInput/DpInput.vue
```diff
@@ -58,9 +58,9 @@ export default {
      * you want to override the label.
      */
     ariaLabelledby: {
-      type: [String, Boolean],
+      type: [String, Boolean, null],
       required: false,
-      default: false
+      default: null
```
salisdemos, line 63: if You change this to `null`, do we still need the boolean in `type` ?

salisdemos (approved):
just one tiny remark

src/components/DpInput/DpInput.vue
hwiem: Good point! 180d240ae93b29e0d697100668095301dd45c326

PR #648, opened 2023-11-21
bug: Align text for accordion button left

gruenbergerdemos:
If the text doesn't fit in one line, the text is aligned center and it does not match the other buttons.

accordion-button


spiess-demos (approved):
lgtm, code review only.

PR #650, opened 2023-11-24
refactor(refs T19759): make tabs work [vue 3 upgrade]

hwiem:
**Integration PR:** #575
**Ticket:** https://yaits.demos-deutschland.de/T19759

- In order to make tabs work again in vue 3, the usage of `$children` in `DpTabs.vue` needs to be removed:
  - $children has been removed in vue 3, see https://v3-migration.vuejs.org/breaking-changes/children.html
  - child components in slots can now be accessed via `useSlots()`
  - in the parent component of `DpTabs` (i.e. where it is used in demosplan), a handler for the change event needs to be added now to update the `activeTabId`; the `change` event is emitted when a tab is clicked and when
setting the initial tab
  - the handler gets the tab id from the change event, then this id can be used to determine the value that needs to be passed to the new `isActive` prop of `DpTab` (by comparing it to the `id`)
- `vue-multiselect` is updated to a version that is compatible with vue3 (but apparently still in beta)

### Related PRs:
- [x] #643 (needs to be merged first)
- [ ] https://github.com/demos-europe/demosplan-core/pull/2365 needs to be merged afterwards


src/components/DpTabs/DpTabs.vue
```diff
@@ -6,151 +6,135 @@
...
+  if (isTabId(hash)) {
+    setActiveTab(hash)
+  }
+}
 
-    if (this.useUrlFragment) {
-      window.addEventListener('hashchange', this.handleHashChange)
-    }
-  },
+const handleTabChange = (id) => {
```
salisdemos, line 83: A general Style question: Do we wan't to omit/use the braces if possible? Do we wan't to standardize that?

hwiem: That's a good question. On the one hand, I'd say we should agree on a consistent solution, on the other hand I find omitting the braces here more confusing than in other cases (e.g. `.map()`). What do you think?

spiess-demos: It is a question of "code reading habit". With `const a = (b) => c`, the "visual signifier" for a function definition was the  parentheses surrounding one or more variable names. Now with `const a = b => c` the arrow is the new visual signifier, and the eye must jump back to the left to look up the arg of the function. My gut feel is that `const a = b => c` is less readable.

spiess-demos (approved):
lgtm (code review only).

src/components/DpTabs/DpTabs.vue
salisdemos: ok. Than we stick with this notation here.
```
a = (x) => foo
```
but I would say, if the arrow function is a callback like in `.map()` we omit them:
```
a.map(x => foo)
```
Did I get it right?

hwiem: At least that's my opinion. :slightly_smiling_face: We could also have a FE vote?

PR #668, opened 2023-12-06
T20094 Add spacing tokens to tailwind

spiess-demos:
Tickets:
- https://yaits.demos-deutschland.de/T20094
- https://yaits.demos-deutschland.de/T30974

This PR **consolidates spacing tokens** and makes them available to Tailwind.

Basically, the already existing space values that proved to be useful are preserved (but renamed in a tailwind-like manner), the larger steps are recalibrated using a geometric progression.

The weird factor of `1.371` is chosen because with this factor, `w-14` turns out to be nearly 600px and there is a smooth transition from the linear progression in steps 1-8.

image

This PR also brings a jsx Table component to Storybook, along with a babel preset-react. This needs to happen because markdown tables are somewhat broken in SB7. All token doc tables are equipped with that new table, now.

:bulb: The `$inuit-base-spacing-unit--huge` class, which is `$inuit-base-spacing-unit * 4` = 96px, is somewhat seldom used in demosplan-core. Therefore, and to be able to start with a geometric progression after 48px, i'd suggest to drop that variable altogether, and replace it with the new `$dp-space-8` variable. The same recalibration will happen with the `width` and `height` variables that are currently defined in demosplan-core scss.


tokens/tokens.space.stories.mdx
```diff
@@ -0,0 +1,84 @@
...
+.huge-element {
+    width: $dp-space-14;
+}
+```
+
+## Usage example - Tailwind utility class
+
+As Tailwind is configured with these variables, too, they can be used as tailwind utility classes.
+They replace the default classes mentioned in the [Tailwind documentation](https://tailwindcss.com/docs/customizing-spacing).
+Instead, the custom spacing scale shown here spacing scale is inherited by the padding, margin, width, height, maxHeight,
```
hwiem, line 79:
```suggestion
Instead, the custom spacing scale shown here is inherited by the padding, margin, width, height, maxHeight,
```

PR #674, opened 2023-12-08
chore: use Tailwind class instead of demosplan-core class

spiess-demos:
As we're already using the `rounded...` classes (and in preparation of removing legacy utility css), let's use Tailwind class instead of demosplan-core class here.


muellerdemos (approved):
I think the 1px border-radius difference between tw and the legacy css won't be recognizable so...
lgtm (code review)

spiess-demos:
> I think the 1px border-radius difference between tw and the legacy css won't be recognizable so... lgtm (code review)

this was not on purpose, thanks for looking into it!

muellerdemos (approved):
lgtm

PR #685, opened 2023-12-21
fix(refs T35888): Adjust broken imports

salisdemos:
Since the imports can't be resolved directly from `~/components` atm. The Modals are broken.
Simple fix is to change the imports, like i did here. In a second step, we have to fix the setup, that we can use the imports the way we want to.


sakutademos (approved):
lgtm

PR #686, opened 2023-12-21
chore(refs T34290): **breaking** Always display Messages on api responses

salisdemos:
until now, this has to be handled by the parent system. But to not forget to use it, checkResponse should always be called.

When releasing this, make sure, that in demosplan-core all ocurences  of `checkResponse` are removed.


src/lib/DpApi.js
```diff
@@ -75,10 +75,16 @@ const doRequest = (async ({ method = 'GET', url, data = {}, headers, params }) =
       ? await response.text()
       : null
 
-    return {
+    const returnData = await checkResponse({
       data: content,
-      status: response.status,
       ok: response.ok,
+      status: response.status
+    }, options.messages || null)
```
spiess-demos, line 82: Where does this "options" come from? Seems like it got removed in the signature of `doRequest` in a recent refactoring, and this was not updated accordingly.

spiess-demos (changes requested):

hwiem:
Can this one be closed?

salisdemos:
Outdatend and overruled by https://github.com/demos-europe/demosplan-ui/pull/1316

PR #690, opened 2023-12-28
Fix(T35880): improve handling of non json responses

sakutademos:
**Ticket:** https://yaits.demos-deutschland.de/T35880, https://yaits.demos-deutschland.de/T35884

**Description:** This PR handles cases where the response does not contain content and cannot be parsed as JSON.

In Bug Ticket [T35884](https://yaits.demos-deutschland.de/T35884), @gruenbergerdemos has experienced a 'Bad Request' issue. Method names are case-sensitive and should be in uppercase, so we receive a 400 Bad Request error if they are in lowercase. Fixed it in this PR as well.

PR in [Demosplan](https://github.com/demos-europe/demosplan-core/pull/2493)


src/lib/DpApi.js
```diff
@@ -54,12 +54,20 @@ const doRequest = (async ({ url, method = 'GET', data = {}, params, options = {}
 
   try {
     const response = await fetch(url, payload)
-    const content = await response.json()
+    const contentTypeHeader = response.headers.get('Content-Type')
+    const contentType = contentTypeHeader ? contentTypeHeader.toLowerCase() : ''
+    const content = contentType.includes('json')
+      ? await response.json()
+      : contentType.includes('text')
+      ? await response.text()
```
spiess-demos, line 62: Why is this fallback to `response.text()` introduced?

sakutademos: If the response has a _text_ content-type, is it necessary to return the text? In Axios, it is parsed automatically. Are there cases where we need to return the text, or is it acceptable to return null instead?

spiess-demos: I see... as we do not know which responseType a specific route may produce, indeed we have to support text as well.

spiess-demos (approved):
lgtm (code review only)

PR #693, opened 2024-01-02
Use tailwind utility classes for height and width

spiess-demos:
To be able to drop "height" and "width" custom utilities, we have to use the appropriate tailwind classes.


sakutademos (approved):
lgtm (code review)

PR #695, opened 2024-01-03
fix: set jackspeak to working version

spiess-demos:
there seems to be a bug in yarn 1 that leads to the "string-width error"
in some cases. while upgrading to yarn 3 is the recommended fix, setting
"jackspeak" to "2.1.1" via package.json resolutions is the next best fix.

see https://github.com/storybookjs/storybook/issues/22431#issuecomment-1630086092


spiess-demos:
this seems to not fix the bug when publishing a package

PR #703, opened 2024-01-09
bug (refs T35953): Don't stringify form data

gruenbergerdemos:
Ticket: https://yaits.demos-deutschland.de/T35953

If you stringify a form data object you get an empty object. For form post we need to send the form data as it is.


hwiem (approved):
lgtm + works, thanks for the fix!

PR #706, opened 2024-01-09
Send uppercase request headers in DpApi

spiess-demos:
Whether it is required by spec is debated, especially when it comes to http2, but as the vuex-json-api plugin also sends headers in the format of "First-Char-Uppercase", this is done here, too, for the time being.


hwiem (approved):
lgtm, thanks! (code review only)

PR #729, opened 2024-01-30
fix(refs T36155): fix the un-/anonymization process

spiess-demos:
**Ticket:** https://yaits.demos-deutschland.de/T36155

**Description:** This PR fixes the issue with the anonymize text function.
- fix the css issue: use the correct classes, as the bubble menu was displayed behind the editor.
- fix the un-anonymization process: to properly un-anonymize text, the attributes (title) should precede the classes in the HTML. This ensures that the anonymized segment can be correctly identified (setValue - line: 91)
- position the bubble-menu above the text


gruenbergerdemos:
(Don't forget to update the changelog)

PR #747, opened 2024-02-13
fix: add clearOnSelect prop to DpMultiselect

spiess-demos:
This change should be released as v0.3.5.1 and picked back into main afterwards.


hwiem (approved):
lgtm, code review only

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ### Fixed
 
 - ([#748](https://github.com/demos-europe/demosplan-ui/pull/748)) Fix required for multiselect ([@gruenbergerdemos](https://github.com/gruenbergerdemos))
+- ([#747](https://github.com/demos-europe/demosplan-ui/pull/747)) Add missing prop clearOnSelect to DpMultiselect (v0.3.5-1) ([@spiess-demos](https://github.com/spiess-demos))
```
hwiem, line 10: I find the duplicate entry a little confusing, but I guess this is how it's normally done? Does it make sense to add that we're including a hotfix?

spiess-demos: as we're including the fix in the upcoming version, i thought that it would be a good idea to also duplicate the entry. i tweaked the changelog entry a little bit in 5e10607ce68b460cf556580865f8cf345a12b48c

PR #777, opened 2024-03-07
fix: convert string to boolean

hwiem:
to make the if statement work, both conditions need to be boolean; otherwise, the warning is logged when `imageButton` is `true` and `tusEndpoint` is set (i.e. != empty string)


spiess-demos (approved):
lgtm (core review).
when introducing the XOR operator at this place, i thought that maybe someone would have to look it up again... funny that today, i was this person.

PR #779, opened 2024-03-11
chore: remove "build:css" script

spiess-demos:
As we do not need it here (Storybook consumes the index.css directly), and also do not need it in projects using demosplan-ui (as they import the tailwind.config and tokens, rather than a compiled file), we can remove the compile step here.


muellerdemos (approved):
lgtm (code review)

PR #784, opened 2024-03-14
Always serialize request params for GET request in dpApi

spiess-demos:
When moving from axios to fetch, it was overlooked that axios serializes params by default, and we only do that occasionally. This leads to the params being dropped in the fetch request whenever "serialize: true" is set via the options arg.

To overcome this, and as we always need to serialize params within get requests, the request params are always appended as stringified value to the url. As the "url" argument may already have params appended, this needs to be handled as well.

The "payload" variable has been renamed into "fetchOptions" to level up readability while distinguishing it from the "options" variable still being used in dpApi.

Also, "options" and "params" has been removed from the request payload as the fetch method does not use these values.

See https://developer.mozilla.org/en-US/docs/Web/API/fetch

### DIFF
--- a/src/lib/DpApi.js
+++ b/src/lib/DpApi.js
@@ -39,28 +39,34 @@ const getHeaders = function ({ headers, url }) {
   }
 }
 
+const appendSerializedUrlParams = (url, params) => {
+  if (!params || Object.keys(params).length === 0) {
+    return url
+  }
+  params = stringify(params, { encodeValuesOnly: true, arrayFormat: 'brackets' })
+
+  // Url params may be already appended before passing it to dpApi, this must be handled accordingly.
+  return url.includes('?') ? `${url}&${params}` : `${url}?${params}`
+}
+
 const doRequest = (async ({ url, method = 'GET', data = {}, params, options = {} }) => {
-  const payload = {
-    method,
-    options,
-    params,
-    headers: getHeaders({ ...params, url })
+  const fetchOptions = {
+    headers: getHeaders({ ...params, url }),
+    method
   }
 
   if (method.toUpperCase() !== 'GET') {
     if (data instanceof FormData) {
-      payload.body = data
+      fetchOptions.body = data
     } else {
-      payload.body = JSON.stringify(data)
+      fetchOptions.body = JSON.stringify(data)
     }
-  } else if (options.serialize === true) {
-    delete payload.options.serialize
-
-    url = `${url}?${stringify(params, { encodeValuesOnly: true, arrayFormat: 'brackets' })}`
+  } else {
+    url = appendSerializedUrlParams(url, params)
   }
 
   try {
-    const response = await fetch(url, payload)
+    const response = await fetch(url, fetchOptions)
     const contentTypeHeader = response.headers.get('Content-Type')
     const contentType = contentTypeHeader ? contentTypeHeader.toLowerCase() : ''
     const content = contentType.includes('json')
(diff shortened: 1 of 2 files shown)


src/lib/DpApi.js
```diff
@@ -39,28 +39,34 @@ const getHeaders = function ({ headers, url }) {
...
+}
+
 const doRequest = (async ({ url, method = 'GET', data = {}, params, options = {} }) => {
-  const payload = {
-    method,
-    options,
-    params,
-    headers: getHeaders({ ...params, url })
+  const fetchOptions = {
+    headers: getHeaders({ ...params, url }),
```
salisdemos, line 54: Its confusing.
Shouldn't this be something like
```suggestion
    headers: getHeaders({ headers, url }),
```
and we have to  pass `headers` as another param to the method?
I know it was like that before. I just wonder how it ever worked. :see_no_evil:

src/lib/DpApi.js
```diff
@@ -39,28 +39,34 @@ const getHeaders = function ({ headers, url }) {
...
+    method
   }
 
   if (method.toUpperCase() !== 'GET') {
     if (data instanceof FormData) {
-      payload.body = data
+      fetchOptions.body = data
```
salisdemos, line 60: why did You rename payload? at least, I think `fetchOptions` don't clarify what it is. Especially because `doRequest` does more than fetch ;-)

salisdemos (changes requested):

src/lib/DpApi.js
spiess-demos: it is a little bit confusing but what actually happens here is that all properties of `headers` are being spread into the object that is passed to `getHeaders`. otherwise you'd have to do something like

```
const { headers } = params
// or: const headers = params?.headers
headers: getHeaders({ headers, url })
```

and within `getHeaders` the object is destructured again, defining only the needed variables:

```
const getHeaders = function ({ headers, url }) {
```

salisdemos: But If You have a look where we pass Headers, they are not within params, but next to them ;-) 
like so:
```
      return dpApi({
        method: 'PATCH',
        url: 'someUrl',
        headers: {
          'Content-type': 'application/vnd.api+json',
          Accept: 'application/vnd.api+json'
        }
```

src/lib/DpApi.js
hwiem: I agree that `fetchOptions` is slightly confusing (because it might also be a function name), but `payload` is also not a very descriptive name. We could change it to `settings` or `config`?

spiess-demos: The variable is passed as the second argument (the "options" argument) into the [`fetch()` global function](https://developer.mozilla.org/en-US/docs/Web/API/fetch#syntax). The decision to name that function "fetch", although it can send all kinds of requests, has been made by some other guys and i only intended to express "these are the options being passed into the fetch function" here.

salisdemos (approved):

hwiem (approved):

src/lib/DpApi.js
spiess-demos: > I just wonder how it ever worked. 🙈

+1. i think that it did not, after dropping axios. It looks like the `headers` property in your last example (when using `dpApi` directly without the method aliases) was just ignored, as it is not forwarded to doRequest. as the api2defaultHeaders contain the correct headers anyways, this is just crucial for "a handful of 1.0 routes" that set custom headers.

PR #785, opened 2024-03-15
fix(refs T36651): display video player controls

hwiem:
without the controls, users can't play videos


salisdemos (approved):
lgtm

spiess-demos (commented):
i guess the dynamic import of the plyr lib is broken, and placing the `controls` attr there just activates the controls of the html5 fallback video element.

PR #786, opened 2024-03-15
fix(refs T36651): make plyr import work

hwiem:
Replaces https://github.com/demos-europe/demosplan-ui/pull/785 which was not an actual fix


spiess-demos (approved):
so here goes the last dynamic import... we should approach tree shaking sooner or later, so that those 117kb will only be included whenever DpVidepPlayer is actually used.
lgtm (code review)

PR #803, opened 2024-03-21
Cleanup dpApi

spiess-demos:
This tackles some leftovers from the axios removal. While options were passed into axios by default before, we do not use it anymore - it was only used to indicate serialization of params which is done by default for get requests now.

- remove "serialize: true" from get route in getFileIdsByHash as params are now always serialized
- properties within the object passed to doRequest as param are reordered to match the most common way of ordering them when calling dpApi() directly in demosplan-core: method first, then url
- add 'headers' property in the object passed to doRequest as param directly instead of wrapping it in params. however this is only used when calling dpApi() directly, as the alias calls like dpApi.post() are used by API2.0 routes which are equipped with api2defaultHeaders anyway.

Linked PR https://github.com/demos-europe/demosplan-core/pull/2878

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


salisdemos (approved):
Thanks for the effort of cleaning this.

PR #804, opened 2024-03-21
fix(refs T36875): add missing procedureId

hwiem:
Ticket: https://yaits.demos-deutschland.de/T36875

Since we are passing a second parameter to `getFileByHash` besides the `hash` (the `procedureId`), we need to provide it in the component that calls the function.

I don't think that this is a good solution because we don't want `procedureId` props in demosplan-ui (too demosplan-specific). However I couldn't find a better solution in the available time. If anyone does, please implement it. :pray: If not, we should consider this a hotfix and find a better solution soon.


salisdemos:
Please don't merge before discussed if https://github.com/demos-europe/demosplan-core/pull/2879 is a better solution

spiess-demos (commented):
lets go with the other solution, which keeps the business logic out of demosplan-ui.

salisdemos:
Overruled by https://github.com/demos-europe/demosplan-core/pull/2879

PR #829, opened 2024-04-12
chore (refs DPLAN-11413): adjust DpCard Styling.

ahmad-demos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11413
https://demosdeutschland.slack.com/archives/C5BA8ABB2/p1712828931605029?thread_ts=1712665548.061619&cid=C5BA8ABB2

- adjust DpCard Styling.

Before:
DpContextualHelp1

After:
DpContextualHelp2

### DIFF
--- a/src/components/DpCard/DpCard.vue
+++ b/src/components/DpCard/DpCard.vue
@@ -8,6 +8,7 @@
       </span>
       <dp-contextual-help
         v-if="headingTooltip !== ''"
+        class="inline-block"
         :text="headingTooltip" />
     </h4><!--
  --><slot />
--- a/src/components/DpCard/DpCard.vue
+++ b/src/components/DpCard/DpCard.vue
@@ -8,7 +8,6 @@
       </span>
       <dp-contextual-help
         v-if="headingTooltip !== ''"
-        class="inline-block"
         :text="headingTooltip" />
     </h4><!--
  --><slot />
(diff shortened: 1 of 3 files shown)


src/components/DpCard/DpCard.vue
```diff
@@ -8,6 +8,7 @@
       </span>
       <dp-contextual-help
         v-if="headingTooltip !== ''"
+        class="inline-block"
```
spiess-demos, line 11: I'd suggest to apply this change directly in [DpContextualHelp](https://github.com/demos-europe/demosplan-ui/blob/79705eaee1c4dfbbcf86b672d0bee95954c0b15d/src/components/DpContextualHelp/DpContextualHelp.vue#L4). In almost all cases when DpContextualHelp is used, we do not want the tooltip icon to be rendered on a new line, but be positioned next to the element it is explaining.

spiess-demos (changes requested):

src/components/DpCard/DpCard.vue
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/829/commits/c813ef8db669786a38fdfe10f51f825fc24e35f7

spiess-demos (approved):
lgtm!

PR #835, opened 2024-04-23
Feat(dplan-11462): DataTable: save width to the session storage

sakutademos:
**Ticket:** [DPLAN-11462](https://demoseurope.youtrack.cloud/issue/DPLAN-11462/macht-spalten-in-der-Abschnittsliste-resizable)

**Description:** This PR adds functionality to save the width of the columns to session storage, eliminating the need to adjust the width every time the user reopens the page. Additionally, it fixes some problems with the `is-resizable` feature.

It saves in the sessionStorage with the key `dpDataTable:data-col-field=${field}`

[PR in Core](https://github.com/demos-europe/demosplan-core/pull/3003)

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -471,9 +471,9 @@ export default {
       if (this.isResizable) {
         this.$nextTick(() => {
           const firstRow = this.tableEl.getElementsByTagName('tr')[0]
-          const tableHeaders = firstRow ? firstRow.children : null
+          const tableHeaderElements = firstRow ? firstRow.children : null
 
-          this.setColsWidth(tableHeaders)
+          this.setColsWidth(tableHeaderElements)
         })
       }
     },
@@ -537,14 +537,14 @@ export default {
     },
 
     setColsWidth (headers) {
-      headers.forEach(tableHeader => {
+      Array.from(headers).forEach(tableHeaderEl => {
         /**
          * Some of childNodes of the first table row are not Element nodes but comments or text.
          * This originates in the Vue template compiler leaving empty html comments when rendering
          * falsy `v-if` blocks. We allow only nodeType "Element" to access its `getBoundingClientRect` api.
          */
-        if(tableHeader.nodeType === 1) {
-          const headerField = tableHeader.getAttribute('data-col-field')
+        if (tableHeaderEl.nodeType === 1) {
+          const headerField = tableHeaderEl.getAttribute('data-col-field')
           const savedColWidth = sessionStorage.getItem(`data-col-field=${headerField}`)
           const fixedWidth = this.getFixedColWidth(headerField)  // some columns, such as 'flyout' and 'wrap', should not be resizable; their width is fixed; the getBoundingClientRect() function should not be applied to them
           const headerFieldWidth = this.getColWidthFromHeaderField(headerField)
@@ -552,9 +552,9 @@ export default {
           const width = fixedWidth
               || savedColWidth
               || headerFieldWidth
-              || `${tableHeader.getBoundingClientRect().width}px`
+              || `${tableHeaderEl.getBoundingClientRect().width}px`
 
-          tableHeader.style.width = width
+          tableHeaderEl.style.width = width
           this.updateSessionStorage(headerField, width)
         }
       })
@@ -650,23 +650,21 @@ export default {
      */
     if (this.isResizable || this.isTruncatable) {
       const firstRow = this.tableEl.getElementsByTagName('tr')[0]
-      const tableHeaders = firstRow ? firstRow.children : null
+      const tableHeaderElements = firstRow ? firstRow.children : null
 
-      this.setColsWidth(tableHeaders)
+      this.setColsWidth(tableHeaderElements)
 
       this.tableEl.style.tableLayout = 'fixed'
       this.tableEl.classList.add('is-fixed')
 
       // Remove styles set by initialMaxWidth and initialWidth after copying rendered width into th styles
       if (this.isResizable) {
-        Array.from(tableHeaders).forEach(tableRow => {
-          Array.from(tableRow).forEach(cell => {
-            if (cell.firstChild) {
-              cell.firstChild.style.width = null
-              cell.firstChild.style.maxWidth = null
-              cell.firstChild.style.minWidth = null
-            }
-          })
+        Array.from(tableHeaderElements).forEach(th => {
+          if (th.firstChild) {
+            th.firstChild.style.width = null
+            th.firstChild.style.maxWidth = null
+            th.firstChild.style.minWidth = null
+          }
         })
       }
     }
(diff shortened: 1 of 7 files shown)


src/components/DpDataTable/DpDataTable.vue
```diff
@@ -597,31 +649,23 @@ export default {
...
-         * This originates in the Vue template compiler leaving empty html comments when rendering
-         * falsy `v-if` blocks. We allow only nodeType "Element" to access its `getBoundingClientRect` api.
-         */
-        if(tableHeader.nodeType === 1) {
-          const width = tableHeader.getBoundingClientRect().width
-          tableHeader.style.width = width + 'px'
-        }
-      })
+      const firstRow = this.tableEl.getElementsByTagName('tr')[0]
+      const tableHeaders = firstRow ? firstRow.children : null
```
spiess-demos, line 653: Nitpick on wording: even if you did not introduce the variable name `tableHeaders`... when reading through the code, it seems unintuitive to name it this way (with "tableHeader" i associate the complete header of the table, not the `th` element actually refered here). Maybe we could name the variable `tableHeaderElements`? idk...

src/components/DpDataTable/DpDataTable.vue
```diff
@@ -597,31 +649,23 @@ export default {
...
 
       // Remove styles set by initialMaxWidth and initialWidth after copying rendered width into th styles
       if (this.isResizable) {
-        const tableRows = Array.from(this.tableEl.children[1].children)
-        tableRows.forEach(tableRow => {
-          Array.from(tableRow.children).forEach(cell => {
-            cell.firstChild.style.width = null
-            cell.firstChild.style.maxWidth = null
-            cell.firstChild.style.minWidth = null
+        Array.from(tableHeaders).forEach(tableRow => {
```
spiess-demos, line 662: I don't understand what is achieved here. `tableHeaders` holds a dom nodeList of all `th` elements, so (with the above naming tweak) it could read...

```suggestion
        Array.from(tableHeaderElement).forEach(th => {
```

So within the outer forEach, the `th` variable should hold a dom node element, not an array. Why are you applying a nested loop here?

src/components/DpDataTable/DpResizableColumn.vue
```diff
@@ -100,6 +118,10 @@ export default {
       document.querySelector('body').removeEventListener('mousemove', this.namedFunc)
       document.querySelector('body').removeEventListener('mouseup', this.stopResize)
       document.querySelector('body').classList.remove('resizing')
+    },
+
+    updateSessionStorage (field, width) {
+      sessionStorage.setItem(`data-col-field=${field}`, width)
```
spiess-demos, line 124: I'd suggest to prefix the key with `dpDataTable:` for better readability when looking through a bunch of storage items in the developer tools.

src/components/DpDataTable/DpResizableColumn.vue
```diff
@@ -71,8 +78,12 @@ export default {
       this.cursorStart = e.pageX
       const resizeBound = this.resize.getBoundingClientRect()
       this.resizeWidth = resizeBound.width
-      const nextBound = this.nextEl.getBoundingClientRect()
-      this.nextWidth = nextBound.width
+
+      if (this.nextEl) {
```
spiess-demos, line 82: Why is it needed to reach out for the next resizable element here?

sakutademos: if the current element's width has been changed, the width of the next element will be calculated and changed as well. I check if the `nextEl` exist, because it returns an `null` when the column is the last in the table. Otherwise, `getBoundingClientRect()` returns an issue.

src/components/DpDataTable/DpDataTable.vue
sakutademos: yes, you are right. I did not change here the logic that was applied before. Thanks for finding this issue. done here https://github.com/demos-europe/demosplan-ui/pull/835/commits/19afd6dc9b016b75745ba7a70f1425433ee8a3d7

src/components/DpDataTable/DpDataTable.vue
sakutademos: right, done here https://github.com/demos-europe/demosplan-ui/pull/835/commits/19afd6dc9b016b75745ba7a70f1425433ee8a3d7

src/components/DpDataTable/DpResizableColumn.vue
sakutademos: Do you mean: `data-col-field=dpDataTable:${field}` ?

src/components/DpDataTable/DpDataTable.vue
```diff
@@ -514,6 +536,30 @@ export default {
...
+      Array.from(headers).forEach(tableHeaderEl => {
+        /**
+         * Some of childNodes of the first table row are not Element nodes but comments or text.
+         * This originates in the Vue template compiler leaving empty html comments when rendering
+         * falsy `v-if` blocks. We allow only nodeType "Element" to access its `getBoundingClientRect` api.
+         */
+        if (tableHeaderEl.nodeType === 1) {
+          const headerField = tableHeaderEl.getAttribute('data-col-field')
+          const savedColWidth = sessionStorage.getItem(`data-col-field=${headerField}`)
+          const fixedWidth = this.getFixedColWidth(headerField)  // some columns, such as 'flyout' and 'wrap', should not be resizable; their width is fixed; the getBoundingClientRect() function should not be applied to them
```
spiess-demos, line 549: Nitpick: this comment could as well live in a `/**` block above the function definition of `getFixedColWidth`.

src/components/DpDataTable/DpResizableColumn.vue
spiess-demos: no, i thought of `dpDataTable:data-col-field=${field}` but i think we do not have any sufficient rules for naming sessionStorage entries at all, we can always discuss this later...

however i find it a little confusing that there the `updateSessionStorage` method is defined in two places but with slightly differing implementations.

src/components/DpDataTable/DpDataTable.vue
```diff
@@ -568,6 +614,12 @@ export default {
       }, {})
 
       this.allWrapped = status
+    },
+
+    updateSessionStorage (storageName, value) {
```
spiess-demos, line 619: `storageName` is not set anywhere.

src/components/DpDataTable/DpDataTable.vue
sakutademos: done here https://github.com/demos-europe/demosplan-ui/pull/835/commits/eda5a1c80fa38a01472b3b0964ee94db7da2fe9f

src/components/DpDataTable/DpDataTable.vue
sakutademos: what do you mean? It is an argument in the func, should it be set somewhere?

spiess-demos (approved):
lgtm! (code review only)

src/components/DpDataTable/DpDataTable.vue
spiess-demos: my bad, did not look closely enough :eyeglasses:

PR #841, opened 2024-04-30
Fix(dplan 1965): improve error validation of the Datepicker

sakutademos:
**Ticket:** [DPLAN-1965](https://demoseurope.youtrack.cloud/issue/DPLAN-1965/Invalid-Inputs-beim-Verfahrenerstellen-werfen-nicht-die-richtigen-Fehlernachrichten) 

**Description:**  This PR adjusts the validation of the Datepicker.

- add the `data-dp-validate-error-fieldname` attribute to the Datepicker via `setAttribute() `to facilitate the display of the field name in case of an error.
- include `name` attributes `r_startdate` and `r_enddate` in the validation form, as the date field need to be validated every time the value changes and not only if the field has an `required` attribute
- an empty string is accepted if the `required` attribute is not present.

[PR in Core](https://github.com/demos-europe/demosplan-core/pull/3038)

### DIFF
--- a/src/components/DpDatepicker/DpDatepicker.vue
+++ b/src/components/DpDatepicker/DpDatepicker.vue
@@ -151,8 +151,8 @@ export default {
     addErrorFieldnameAttribute () {
       const datePickerInput = document.getElementsByName(this.name)[0]
       /**
-       * To ensure proper error handling, this attribute needs to be added every time the Datepicker is mounted and updated.
-       * This attribute is needed by the validation to display the field name in case of an error
+       * This attribute is needed for validation to display the field name in case of an error
+       * and must be set every time the Datepicker is mounted or updated.
        */
       datePickerInput.setAttribute('data-dp-validate-error-fieldname', this.dataDpValidateErrorFieldname)
     },
--- a/src/components/DpDatepicker/DpDatepicker.vue
+++ b/src/components/DpDatepicker/DpDatepicker.vue
@@ -164,6 +164,7 @@ export default {
   },
 
   mounted () {
+    // https://github.com/mathislucka/a11y-datepicker
     const config = {
       ...this.calendarsAfter > 0 ? { monthsAfterCurrent: this.calendarsAfter } : {},
       ...this.calendarsBefore > 0 ? { monthsBeforeCurrent: this.calendarsBefore } : {},
--- a/src/components/DpDatepicker/DpDatepicker.vue
+++ b/src/components/DpDatepicker/DpDatepicker.vue
@@ -142,8 +142,20 @@ export default {
   },
 
   methods: {
+    isValidDate (dateString) {
+      // Attempt to create a Date object from the given string
+      const date = new Date(dateString);
+
+      // Check if the created date is valid
+      // Also, check if the input string is not 'Invalid Date'
+      return !isNaN(date.getTime()) && date.toString() !== 'Invalid Date';
+    },
+
     emitUpdate (e) {
       const currentVal = e.target.value
+      const isValid = this.isValidDate(currentVal)
+      console.log('isValid', isValid)
+      console.log('currectVal', currentVal)
       const date = this.datepicker.getDateAsString()
       const valueToEmit = date === currentVal ? date : currentVal
       this.$emit('input', valueToEmit)
--- a/src/components/DpDatepicker/DpDatepicker.vue
+++ b/src/components/DpDatepicker/DpDatepicker.vue
@@ -150,7 +150,11 @@ export default {
   methods: {
     addErrorFieldnameAttribute () {
       const datePickerInput = document.getElementsByName(this.name)[0]
-      datePickerInput.setAttribute('data-dp-validate-error-fieldname', this.dataDpValidateErrorFieldname) // this attribute is required by the validation to showcase the field name in case of an error
+      /**
+       * To ensure proper error handling, this attribute needs to be added every time the Datepicker is mounted and updated.
+       * This attribute is needed by the validation to display the field name in case of an error
+       */
+      datePickerInput.setAttribute('data-dp-validate-error-fieldname', this.dataDpValidateErrorFieldname)
     },
 
     emitUpdate (e) {
--- a/src/components/DpDatepicker/DpDatepicker.vue
+++ b/src/components/DpDatepicker/DpDatepicker.vue
@@ -31,6 +31,12 @@ export default {
       default: 'datepicker'
     },
 
+    dataDpValidateErrorFieldname: {
+      type: String,
+      required: false,
+      default: ''
+    },
+
     disabled: {
       type: Boolean,
       required: false,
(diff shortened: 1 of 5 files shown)


src/lib/validation/utils/validateDatepicker.js
```diff
@@ -2,11 +2,19 @@ import { toggleErrorClass } from './helpers'
 
 export default function validateDatepicker (input) {
   let isValid = true
+
   if (input.value === '') {
-    isValid = false
+    isValid = !input.hasAttribute('required')
```
spiess-demos, line 7: I don't understand this logic. for me this means, setting isValid to true for all instances where no input is required. But, i can also input invalid data into a non-required field!

src/lib/validation/utils/helpers.js
```diff
@@ -2,7 +2,7 @@ import getScrollTop from '../../../utils/getScrollTop'
 import prefixClass from '../../../utils/prefixClass'
 
 function getAllInputsArray (form) {
-  return Array.from(form.querySelectorAll('[pattern], [required], input[type="email"], .tiptap__input--hidden.is-required, .multiselect.is-required, fieldset.is-required, [minlength], [maxlength], [data-dp-validate-maxlength], [data-dp-validate-should-equal]'))
+  return Array.from(form.querySelectorAll('[pattern], [required], input[type="email"], input[name="r_startdate"], input[name="r_enddate"], .tiptap__input--hidden.is-required, .multiselect.is-required, fieldset.is-required, [minlength], [maxlength], [data-dp-validate-maxlength], [data-dp-validate-should-equal]'))
```
spiess-demos, line 5: I'd opt against hardcoding business logic from demosplan-core here. Instead, we could always validate date inputs with ```[data-ad-id]``` (which is produced by the datepicker lib).

src/lib/validation/utils/helpers.js
```diff
@@ -16,6 +16,8 @@ function shouldValidate (input) {
     isRequired ||
     input.hasAttribute('pattern') ||
     input.getAttribute('type') === 'email' ||
+    input.getAttribute('name') === 'r_startdate' ||
+    input.getAttribute('name') === 'r_enddate' ||
```
spiess-demos, line 20: see the other comment. let's change this into:

```suggestion
    input.hasAttribute('data-ad-id')
```

and always validate date fields.

src/lib/validation/utils/validateDatepicker.js
```diff
@@ -2,11 +2,19 @@ import { toggleErrorClass } from './helpers'
...
-    const regex = /^(?:(?:31(.)(?:0?[13578]|1[02]))\1|(?:(?:29|30)(.)(?:0?[13-9]|1[0-2])\2))(?:(?:1[6-9]|[2-9]\d)?\d{2})$|^(?:29(.)0?2\3(?:(?:(?:1[6-9]|[2-9]\d)?(?:0[48]|[2468][048]|[13579][26])|(?:(?:16|[2468][048]|[3579][26])00))))$|^(?:0?[1-9]|1\d|2[0-8])(.)(?:(?:0?[1-9])|(?:1[0-2]))\4(?:(?:1[6-9]|[2-9]\d)?\d{2})$/
+    /**
+     * DpDatePicker accepts the dateFormat: 'dd.mm.yyyy'
+     * Regex to check date from https://www.regextester.com/97612
+     * Allows dates following these rules:
+     * - days may have leading zeros. 1-31. max 2 digits
+     * - months may have leading zeros. 1-12. max 2 digits
+     * - years 1900-2099. 4 digits
+     */
+    const regex = /^\s*(3[01]|[12][0-9]|0?[1-9])\.(1[012]|0?[1-9])\.((?:19|20)\d{2})\s*$/
```
spiess-demos, line 17: While narrowing down the possible input to meet the Datepicker capabilities is fine (reads: no blocker here), i feel like we should possibly allow users to input dates as they like.

Another question: how do we educate users about the expected date format? the datepicker lib does not offer a placeholder, so...

spiess-demos (dismissed):

src/lib/validation/utils/validateDatepicker.js
sakutademos: it is about the empty string, whether the empty string is allowed or not. The problem is that we have to validate the Datepicker every time the user enters smth, not only if this field is required. Similar to the email field, it should be validated every time because there is a pattern. In this case, we can allow an empty string, like [here](https://github.com/demos-europe/demosplan-ui/blob/main/src/lib/validation/utils/validateInputField.js#L36)

src/lib/validation/utils/helpers.js
sakutademos: Hmm, if this is a specific Datepicker attribute, then we can do it. Thanks

src/lib/validation/utils/helpers.js
sakutademos: done

src/components/DpDatepicker/DpDatepicker.vue
```diff
@@ -142,12 +148,22 @@ export default {
   },
 
   methods: {
+    addErrorFieldnameAttribute () {
+      const datePickerInput = document.getElementsByName(this.name)[0]
+      /**
+       * This attribute is needed for validation to display the field name in case of an error
+       * and must be set every time the Datepicker is mounted or updated.
+       */
```
salisdemos, line 156: why does it need an update on every change?

src/lib/validation/utils/helpers.js
```diff
@@ -16,6 +16,7 @@ function shouldValidate (input) {
     isRequired ||
     input.hasAttribute('pattern') ||
     input.getAttribute('type') === 'email' ||
+    input.hasAttribute('data-ad-id') ||
```
salisdemos, line 19: where comes this from? Why is this needed?

src/lib/validation/utils/validateDatepicker.js
salisdemos: A solution could be to reformat the date before passing it to the datepicker. But that would be a story by its own.

src/lib/validation/utils/helpers.js
sakutademos: the Datepicker uses this specific [attribute](https://github.com/mathislucka/a11y-datepicker/blob/master/src/js/draw/Draw.js#L27). See here a comment from @spiess-demos https://github.com/demos-europe/demosplan-ui/pull/841#discussion_r1587279898 . I think it is a good idea to get the Datepicker input with this attribute, but I think I should add a comment for a better readability

src/components/DpDatepicker/DpDatepicker.vue
sakutademos: after updating the input value, the `data-dp-validate-error-fieldname` attribute that was added during mounting disappears from DOM, so I also [include it in the `emitUpdate` action](https://github.com/demos-europe/demosplan-ui/pull/841/files#diff-9749bd9f6c18eb9c907ecf005c35b25dc64bc5431f99f694aba87c0ee5c1c519R166)

salisdemos: Thanks for the explanation

src/lib/validation/utils/helpers.js
salisdemos: Ok. Thanks. The Comment helps a lot

salisdemos (approved):

src/lib/validation/utils/validateDatepicker.js
spiess-demos: thanks for adding the tooltip. for now this is ok. in the long run we should display the expected format permanently, as it could easily missed in a tooltip.

PR #843, opened 2024-05-02
Make length prop validation work

spiess-demos:
As the "length" variable is assigned to "maxlength" prop, without calling it, it does not need to be a function, rather an object.


hwiem (approved):
lgtm, code review only

PR #844, opened 2024-05-03
fix(refs DPLAN-11529): Don't change Width from Next El on Datatable resize

salisdemos:
I removed the behavior in DpDataTable, that on resize the followed Column gets resized too. Now all following cols get "pushed away".

I don't know what GitHub/my IDE did, but I only changed DpResizeableColumn. DpDataTable should only be autolinted.

**Ticket** https://demoseurope.youtrack.cloud/issue/DPLAN-11529/Abschnittstabelle-Nach-Breite-verkleinern-einer-Spalte-wird-die-Spalte-rechts-daneben-sehr-breit-gezogen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
+
+- ([#844](https://github.com/demos-europe/demosplan-ui/pull/844)) Prevent changing Width on following DataTable Element, while resizing ([@salisdemos](https://github.com/salisdemos))
+
 ## v0.3.11 - 2024-05-03
 
 ### Fixed
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -18,83 +18,83 @@
       </colgroup>
 
       <thead>
-        <dp-table-header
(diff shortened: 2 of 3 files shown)


spiess-demos (approved):
Works as expected in cases when the table is larger than its container. If it is not, decreasing the size of a column leads to other columns changing their width, too, which feels a little odd. However this does not need to be fixed within this PR.

salisdemos:
> Works as expected in cases when the table is larger than its container. If it is not, decreasing the size of a column leads to other columns changing their width, too, which feels a little odd. However this does not need to be fixed within this PR.

But what would be Your expected behaviour? Should the table then be less then 100%? I thought about that, but decided to keep the 100% at min.

spiess-demos:
> But what would be Your expected behaviour? Should the table then be less then 100%? I thought about that, but decided to keep the 100% at min.

this is correct, i'd also keep the 100% width. but if i resize a column to be smaller than before, i'd expect the following column to grow (so that would be the same behavior than before).

salisdemos:
Oh. Yeah. I didn't think about that obvious solution. should be done in  follow up. (had that almost like that. It was just somehow buggy)

salisdemos:
> this is correct, i'd also keep the 100% width. but if i resize a column to be smaller than before, i'd expect the following column to grow (so that would be the same behavior than before).

What came to mind: How does the table behaves when the Last Col gets shrunken on a Table smaller than 100% ?

PR #848, opened 2024-05-08
fix(refs DPLAN-1898): Destroy Tooltip instead of hiding

salisdemos:
It was the idea to hide the tooltip on mouseleave and only destroy it after 300ms to prevent recreating it again and again. But it turned out that it then sometimes got created twice. That had the effect, that the second tooltip didn't disappear.

**Ticket** DPLAN-1898

### DIFF
--- a/src/components/DpTooltip/utils/tooltip.js
+++ b/src/components/DpTooltip/utils/tooltip.js
@@ -41,7 +41,7 @@ const initTooltip = (el, value, options) => {
 
   el.setAttribute('aria-describedby', id)
 
-  handleCreateTooltip = () => showTooltip(
+  handleCreateTooltip = () => createTooltip(
     id,
     el,
     options,
@@ -55,7 +55,7 @@ const initTooltip = (el, value, options) => {
   el.addEventListener('blur', handleRemoveTooltip)
 }
 
-const showTooltip = async (id, wrapperEl, { place = 'top', container = 'body', classes = '' }, zIndex)  => {
+const createTooltip = async (id, wrapperEl, { place = 'top', container = 'body', classes = '' }, zIndex)  => {
   if (!document.getElementById(wrapperEl.getAttribute('aria-describedby'))) {
     const value = tooltips[id]
     // this has to be in sync with the Template in DpTooltip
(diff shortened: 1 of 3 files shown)


src/components/DpTooltip/utils/tooltip.js
```diff
@@ -72,7 +63,7 @@ const initTooltip = (el, value, options) => {
     options,
     zIndex
   )
-  handleHideTooltip = () => hideTooltip(document.getElementById(el.getAttribute('aria-describedby')))
+  handleHideTooltip = () => deleteTooltip(document.getElementById(el.getAttribute('aria-describedby')))
```
spiess-demos, line 66: Nitpick: As we're not actually hiding anything here, i'd suggest to unify wording:

```suggestion
  handleRemoveTooltip = () => removeTooltip(document.getElementById(el.getAttribute('aria-describedby')))
```

src/components/DpTooltip/utils/tooltip.js
```diff
@@ -83,8 +74,6 @@ const initTooltip = (el, value, options) => {
 const showTooltip = async (id, wrapperEl, { place = 'top', container = 'body', classes = '' }, zIndex)  => {
   if (!document.getElementById(wrapperEl.getAttribute('aria-describedby'))) {
```
spiess-demos, line 75: This function could be merged into `createTooltip`. This way, the wording would also be more appropriate (as we're not showing anything, just recreating the elements).

spiess-demos (commented):
Nice! Only a few naming suggestions.

src/components/DpTooltip/utils/tooltip.js
salisdemos: like so?
843c1dee22976879111309c39c5ee7e470acbb14

src/components/DpTooltip/utils/tooltip.js
salisdemos: 843c1dee22976879111309c39c5ee7e470acbb14

spiess-demos (approved):
lgtm! (code review)

PR #850, opened 2024-05-14
fix(DPLAN-2231): fix 'data-dp-validate-if' within a fieldset

sakutademos:
**Ticket:** [DPLAN-2231](https://demoseurope.youtrack.cloud/issue/DPLAN-2231/Fehlermeldung-das-Verfahren-nicht-verortet-ist-in-jeden-Assistent-schritt-angezeigt)

**Description:** This PR fixes an issue with validation when a 'data-dp-validate-if' attribute is provided in an input within a fieldset instead of directly within a form, because the query selector (line 20) should search the entire form rather than just the fieldset; otherwise, it returns null.

- check if a `fieldset` is provided instead of a `form` when using `input.closest('[data-dp-validate]')`
- cognitive complexity has been reduced for the 'assignHandlersForSingleInput' method to address a SonarCloud issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Fixed
 
-- ([#850](https://github.com/demos-europe/demosplan-ui/pull/850)) Fix the validation of the relation when the 'data-dp-validate-if' is provided ([@sakutademos](https://github.com/sakutademos))
+- ([#850](https://github.com/demos-europe/demosplan-ui/pull/850)) Input validation: use form as validation container ([@sakutademos](https://github.com/sakutademos))
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+
+- ([#847](https://github.com/demos-europe/demosplan-ui/pull/847)) Tailwind: add "scrollbar-none" class ([@spiess-demos](https://github.com/spiess-demos))
+
 ## v0.3.12 - 2024-05-06
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
+
+- ([#850](https://github.com/demos-europe/demosplan-ui/pull/850)) Fix the validation of the relation when the 'data-dp-validate-if' is provided ([@sakutademos](https://github.com/sakutademos))
+
 ### Added
 
 - ([#847](https://github.com/demos-europe/demosplan-ui/pull/847)) Tailwind: add "scrollbar-none" class ([@spiess-demos](https://github.com/spiess-demos))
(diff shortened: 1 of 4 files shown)


src/lib/validation/utils/assignHandlersForInputs.js
```diff
@@ -21,7 +21,8 @@ function assignHandlersForSingleInput (input) {
       allConditions.forEach(condition => {
         const comparisonType = condition.indexOf('!==') > -1 ? 'isNotEqual' : 'isEqual'
         const matchers = condition.split(comparisonType === 'isNotEqual' ? '!==' : '===')
-        const form = input.closest('[data-dp-validate]')
+        const validationContainer = input.closest('[data-dp-validate]')
+        const form = validationContainer.tagName === 'FIELDSET' && validationContainer.form ? validationContainer.form : validationContainer
```
spiess-demos, line 25: I feel this logic needs a little commenting.

Also, i think we should move the [dp-validate documentation from demosplan-core](https://github.com/demos-europe/demosplan-documentation/blob/e11a15498b1ca8d97f519f2cc750a9914787a0a8/development/application-architecture/validation/frontend/README.md) into demosplan-ui and check if it is outdated (certainly not in this PR).

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ### Fixed
 
 - ([#858](https://github.com/demos-europe/demosplan-ui/pull/858)) DpCheckbox: fix vue warning related to type checking ([@sakutademos](https://github.com/sakutademos))
+- ([#850](https://github.com/demos-europe/demosplan-ui/pull/850)) Fix the validation of the relation when the 'data-dp-validate-if' is provided ([@sakutademos](https://github.com/sakutademos))
```
hwiem, line 14: I think this entry does not express very clearly what is going on. Maybe something like
`Input validation: Always use form as validation container`
or
`Input validation: Always use querySelector on form element`
is slightly better to understand?
(I like to add a context like "Input validation:" at the beginning of the message)

hwiem (commented):
I agree that an explanatory comment would be helpful (what @spiess-demos said).
Code lgtm, just a comment on the changelog entry

CHANGELOG.md
sakutademos: right, done here https://github.com/demos-europe/demosplan-ui/pull/850/commits/ae879da1bed75f3ad12d621fa9801060d9eadc93

src/lib/validation/utils/assignHandlersForInputs.js
sakutademos: add a comment here https://github.com/demos-europe/demosplan-ui/pull/850/commits/ae879da1bed75f3ad12d621fa9801060d9eadc93

sakutademos: We currently have a Storybook in demosplan-ui. What do you mean by _move it into demosplan-ui_? create a docu for demosplan-ui?

hwiem (approved):

src/lib/validation/utils/assignHandlersForInputs.js
spiess-demos: Correctly. At the moment we document only components and tokens via Storybook. Neither "directives", "lib" (which is highly unsorted and should be sorted upfront), "mixins" or "utils" are documented at all. At a second glance i think we should at first decide how to structure util/lib code within demosplan-ui, and only after that document things we decided to be "multi purpose enough" to stay in demosplan-ui.

spiess-demos (approved):
lgtm (code review)

PR #851, opened 2024-05-14
fix(DPLAN-3297): DpLabel: adjust tooltip

sakutademos:
**Ticket:** [DPLAN-3297](https://demoseurope.youtrack.cloud/issue/DPLAN-3297/2-Als-Fachplanung-mochte-ich-Institutionen-per-E-Mail-zu-einem-Verfahren-einladen-konnen-obwohl-sie-noch-nicht-in-BLP)

**Description:** DpLabel: use the DpContextualHelp component instead of the tooltip directive and place it near the label.

### DIFF
--- a/src/components/DpLabel/DpLabel.vue
+++ b/src/components/DpLabel/DpLabel.vue
@@ -74,12 +74,6 @@ export default {
     }
   },
 
-  data () {
-    return {
-      ariaLabel: de.contextualHelp
-    }
-  },
-
   computed: {
     /**
      * List of Hints
--- a/src/components/DpLabel/DpLabel.vue
+++ b/src/components/DpLabel/DpLabel.vue
@@ -14,25 +14,26 @@
           v-cleanhtml="h" />
       </span>
     </span>
-    <i
+    <dp-contextual-help
       v-if="tooltip !== ''"
-      :class="prefixClass('fa fa-question-circle u-mt-0_125 ml-auto')"
+      :class="prefixClass('mt-px ml-0.5')"
       :aria-label="ariaLabel"
-      v-tooltip="tooltip" />
+      :text="tooltip" />
   </label>
 </template>
 
 <script>
-import { CleanHtml, Tooltip } from '~/directives'
+import { CleanHtml } from '~/directives'
 import { de } from '~/components/shared/translations'
 import { prefixClassMixin } from '~/mixins'
+import DpContextualHelp from '~/components/DpContextualHelp/DpContextualHelp.vue'
 
 export default {
   name: 'DpLabel',
+  components: { DpContextualHelp },
 
   directives: {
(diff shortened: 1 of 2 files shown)


src/components/DpLabel/DpLabel.vue
```diff
@@ -14,25 +14,26 @@
           v-cleanhtml="h" />
       </span>
     </span>
-    <i
+    <dp-contextual-help
       v-if="tooltip !== ''"
-      :class="prefixClass('fa fa-question-circle u-mt-0_125 ml-auto')"
+      :class="prefixClass('mt-px ml-0.5')"
       :aria-label="ariaLabel"
```
spiess-demos, line 20: `aria-label` is already set in DpContextualHelp (with the same value as here), so it can be removed here.

src/components/DpLabel/DpLabel.vue
```diff
@@ -14,25 +14,26 @@
...
+      :text="tooltip" />
   </label>
 </template>
 
 <script>
-import { CleanHtml, Tooltip } from '~/directives'
+import { CleanHtml } from '~/directives'
 import { de } from '~/components/shared/translations'
 import { prefixClassMixin } from '~/mixins'
+import DpContextualHelp from '~/components/DpContextualHelp/DpContextualHelp.vue'
```
spiess-demos, line 29: Does it also work to include this with

```suggestion
import DpContextualHelp from '~/components/DpContextualHelp'
```
?

spiess-demos (changes requested):

src/components/DpLabel/DpLabel.vue
sakutademos: https://github.com/demos-europe/demosplan-ui/pull/851/commits/71ef394cfeb9ff8470151ba6a8c3a1d29b4ed80b

src/components/DpLabel/DpLabel.vue
sakutademos: done here https://github.com/demos-europe/demosplan-ui/pull/851/commits/71ef394cfeb9ff8470151ba6a8c3a1d29b4ed80b

spiess-demos: Seems like we can also remove it [in data](https://github.com/demos-europe/demosplan-ui/pull/851/files#diff-65de55f21df8cc95b480ecefa574f35dcf117e8053d810c1a9a13a56ddabb1ccR77-R81), or am i missing something?

sakutademos: no, you are right, done https://github.com/demos-europe/demosplan-ui/pull/851/commits/e35402d5656660f196cab7083fd01081768c71d1

spiess-demos (approved):
Lgtm! Nice to see the old tooltips slowly disappear 💨

PR #860, opened 2024-05-16
bug (refs DPLAN-11472): Only setAttribute if datePickerInput is defined

gruenbergerdemos:
Ticket: [DPLAN-11472](https://demoseurope.youtrack.cloud/issue/DPLAN-11472)

Only set attribute if datePickerInput is defined


sakutademos (approved):
lgtm

PR #865, opened 2024-05-21
bug (refs DPLAN-11586): Remove header from form posts

gruenbergerdemos:
Ticket: [DPLAN-11586](https://demoseurope.youtrack.cloud/issue/DPLAN-11586/Stellungnahme-als-Entwurf-speichern-spater-nicht-sichtbar-unter-Meine-Entwurfe)

If you're sending form data using FormData, you should not manually set the Content-Type header. The browser will automatically set it to multipart/form-data and append the necessary boundary parameter.


spiess-demos (approved):
lgtm (code review)
seems like setting the headers is not needed in this case.

PR #891, opened 2024-06-06
bug(refs DPLAN-2071): Escape special characters for searching

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-2071/Text-fehlt-wenn-die-Verfahrenssuche-ein-enthalt

Escape whitspaces and `+` for searching the table.

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -460,7 +460,7 @@ export default {
       if (this.searchString === null || this.searchString.length < 1) {
         return new RegExp()
       }
-      const searchTerm = this.searchString.replace(/\s*/ig, '\\s*')
+      const searchTerm = this.escapeSpecialCharacters(this.searchString)
       return new RegExp(searchTerm, 'ig')
     },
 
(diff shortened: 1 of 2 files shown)


src/components/DpDataTable/DpDataTable.vue
```diff
@@ -497,6 +497,23 @@ export default {
...
+     * This method is used to escape special characters in a string.
+     * It specifically targets spaces and plus signs.
+     *
+     * @param {string} string - The string to be processed.
+     * @returns {string} - The processed string with special characters escaped.
+     */
+    escapeSpecialCharacters (string) {
+      return string.replace(/[\s+]/g, (match) => {
+        if (match === ' ') {
+          return '\\s*';
```
salisdemos, line 510: I actually don't now what "You" (the person before You) wants to archive with that. Can You give me an example input with the expected output?

salisdemos (changes requested):

src/components/DpDataTable/DpDataTable.vue
gruenbergerdemos: I thought it was used to escape whitespaces eg. `Test test`.

salisdemos: That doesn't make sense to me. I could imagine that we want to replace special whitespaces to regular whitespaces. That would be something like `.replace(/\s/g, ' ')` . A bit the other way round ;-)

salisdemos: if so  i suggest something like
```
 escapeSpecialCharacters (string) {
      return string
        .replace(/(\s)/g, ' ')
        .replace(/([+])/g, '\\$&') // here we can add other special chars if necessary without too much adjustment
```

salisdemos (changes requested):

src/components/DpDataTable/DpDataTable.vue
gruenbergerdemos: Done in [064afd7](https://github.com/demos-europe/demosplan-ui/pull/891/commits/064afd77bc01700fbf76a78d16846774eb4c1b73)

salisdemos (approved):

PR #892, opened 2024-06-10
bug(refs DPLAN-1879): Add hidden attribute for label

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-2759

Sometimes we don't want a label visually, but there should be one nonetheless for accessibility reasons.

Related PR
https://github.com/demos-europe/demosplan-core/pull/3235

### DIFF
--- a/src/components/DpInput/DpInput.vue
+++ b/src/components/DpInput/DpInput.vue
@@ -2,13 +2,13 @@
   <div :class="containerWidth !== '' ? prefixClass(containerWidth) : false">
     <dp-label
       v-if="label.text !== ''"
-      :class="{ 'hide-visually': label.hidden }"
       v-bind="{
         ...label,
         for: id,
         hint: labelHint,
         required: required
-      }" /><!--
+      }"
+      :class="{ 'hide-visually': label.hidden }" /><!--
  --><input
       :id="id"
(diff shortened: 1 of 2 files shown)


src/components/DpInput/DpInput.vue
```diff
@@ -2,6 +2,7 @@
   <div :class="containerWidth !== '' ? prefixClass(containerWidth) : false">
     <dp-label
       v-if="label.text !== ''"
+      :class="{ 'hide-visually': label.hidden }"
```
salisdemos, line 5: class should be below v-bind. Not only for the alphabetic order, but also because with vue3 there will be a change in the handling of the order:
https://v3-migration.vuejs.org/breaking-changes/v-bind#_3-x-syntax

salisdemos (changes requested):

salisdemos (approved):

src/components/DpInput/DpInput.vue
```diff
@@ -133,12 +134,13 @@ export default {
       type: Object,
       default: () => ({
         bold: true,
+        hidden: false,
```
spiess-demos, line 137: this reads as if the whole input field would be hidden.

PR #894, opened 2024-06-11
bug(refs DPLAN-1879): Move hidden prop to DpLabel

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-2759

Hidden prop should rather be handled directly in DpLabel instead of DpInput. 

Related PR
https://github.com/demos-europe/demosplan-ui/pull/892

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 - ([#893](https://github.com/demos-europe/demosplan-ui/pull/893)) Add data attributes for E-2-E Test ([@ahmad-demos](https://github.com/ahmad-demos))
 - ([#892](https://github.com/demos-europe/demosplan-ui/pull/892)) DpInput: Add hidden attribute for label ([@gruenbergerdemos](https://github.com/gruenbergerdemos))
+- ([#894](https://github.com/demos-europe/demosplan-ui/pull/894)) DpLabel: Add hidden attribute for visually hiding the label ([@gruenbergerdemos](https://github.com/gruenbergerdemos))
 
 ## v0.3.17 - 2024-05-30
 
--- a/src/components/DpLabel/DpLabel.vue
+++ b/src/components/DpLabel/DpLabel.vue
@@ -1,6 +1,6 @@
 <template>
   <label
-    :class="prefixClass(['o-form__label flex', bold ? 'weight--bold' : 'weight--normal', hints.length > 0 ? 'has-hint' : ''])"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 - ([#893](https://github.com/demos-europe/demosplan-ui/pull/893)) Add data attributes for E-2-E Test ([@ahmad-demos](https://github.com/ahmad-demos))
 - ([#892](https://github.com/demos-europe/demosplan-ui/pull/892)) DpInput: Add hidden attribute for label ([@gruenbergerdemos](https://github.com/gruenbergerdemos))
```
spiess-demos, line 10: do we still want to persist this entry, then?

src/components/DpLabel/DpLabel.vue
```diff
@@ -49,6 +49,12 @@ export default {
       required: true
     },
 
+    hidden: {
```
spiess-demos, line 52: Also if this is (rather) obvious: a one-line documentation on the prop will later ease storybook completion.

spiess-demos (changes requested):

CHANGELOG.md
gruenbergerdemos: I left it there, because we still added the hidden attribute to the label prop in DpInput. Do you think it is confusing?

src/components/DpLabel/DpLabel.vue
gruenbergerdemos: If we use class `sr-only` should we call the prop also sr-only? :thinking:

CHANGELOG.md
spiess-demos: you're right, technically DpInput inherits the attr as well. so the changelog entry is ok.

src/components/DpLabel/DpLabel.vue
spiess-demos: idk.. i think while `srOnly` or `screenReaderOnly` will better describe the intention of this prop, we may want to strive for consistency here... in the button component the prop reads `hideText`... so as we're hiding the whole label here including hint and tooltip, we may go with just `hide`? idk.

gruenbergerdemos: Done in [023db8c](https://github.com/demos-europe/demosplan-ui/pull/894/commits/023db8c116cc3f07c4b1e7f3f9502bc427b8be53)

spiess-demos (approved):
lgtm! (code review)

PR #900, opened 2024-06-13
refactor: Use tailwind class for hiding visually

gruenbergerdemos:
Refactor: Use tailwind class for hiding visually instead of custom class. 

Related PR https://github.com/demos-europe/demosplan-core/pull/3249


spiess-demos (approved):
lgtm (code review)

PR #902, opened 2024-06-14
Remove !important tailwind config

spiess-demos:
As this setting may be added when extending tailwind.config.js from other projects, this does not need to be set here.

See https://github.com/demos-europe/demosplan-core/pull/3272 for reasoning about this change.


hwiem (approved):
lgtm!

PR #911, opened 2024-06-27
fix: replace the obsolete class with a tailwind class

sakutademos:
**Description:** This PR uses the Tailwind css class `text-right` instead of the outdated `text--right`

The name `align` as a prop is not a good choice as there is a conflict with the HTML attribute `align`. The phpstrom flags it because modern HTML has deprecated this usage. Using `align` as a prop name can cause confusion since it might be interpreted as an HTML attribute.

But I am not sure which naming for the prop is better in this case? 
`textAlign`, `contentAlign`, `alignment`, `position` ? I think that the `text-alignment` is not a bad option, oder? 

image

[PR in core](https://github.com/demos-europe/demosplan-core/pull/3456)


salisdemos (approved):
Thanks for fixing!

sakutademos:
> Thanks for fixing!

There is a question about renaming the align prop :) Should I leave it as it is or rename it?

salisdemos:
> > Thanks for fixing!
> 
> There is a question about renaming the `align` prop :) What is the best option?

good question:
Maybe we can rename it like "buttonAlign" or something like that?!?

src/components/DpButtonRow/DpButtonRow.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div
-    :class="[`text--${align}`, $attrs.class]"
+    :class="[`text-${align}`, $attrs.class]"
```
spiess-demos, line 3: Constructing the classes in such a way will stop tailwind from discovering them when parsing the source files. See https://tailwindcss.com/docs/content-configuration#dynamic-class-names

spiess-demos (changes requested):
One concern.
As for the question on naming: how about `alignment`?

spiess-demos (approved):
lgtm!

PR #931, opened 2024-07-16
chor (refs DPLAN-11885): add Jest Test for the DpAccordion component.

ahmad-demos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11885/Unit-Tests-schreiben

- add Jest Test for the DpAccordion component.

### DIFF
--- a/tests/DpAccordion.spec.js
+++ b/tests/DpAccordion.spec.js
@@ -1,0 +1,58 @@
+import { shallowMount } from '@vue/test-utils'
+import DpAccordion from '~/components/DpAccordion/DpAccordion.vue'
+import DpTransitionExpand from '~/components/DpTransitionExpand'
+
+describe('DpAccordion', () => {
+  let wrapper
+
+  beforeEach(() => {
+    wrapper = shallowMount(DpAccordion, {
+      propsData: {
+        dataCy: 'accordionToggle',
+        fontWeight: 'bold',
+        compressed: false,
+        isOpen: false,
+        title: 'Test Title'
+      }
+    })
+  })
+
+  it('renders component correctly', () => {
+    expect(wrapper.exists()).toBe(true)
+  })
+
+  it('renders with the correct title and default props', () => {
+    expect(wrapper.find('button').text()).toContain('Test Title')
+    expect(wrapper.find('button').classes()).toContain('weight--bold')
+    expect(wrapper.find('button').attributes('aria-expanded')).toBe('false')
+    expect(wrapper.find('button').attributes('data-cy')).toContain('accordionToggle')
+  })
+
+  it('toggles the accordion visibility when the button is clicked', async () => {
+    expect(wrapper.vm.isVisible).toBe(false)
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.vm.isVisible).toBe(true)
+    expect(wrapper.find('button').attributes('aria-expanded')).toBe('true')
+  })
+
+  it('emits an event when the accordion is toggled', async () => {
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.emitted('item:toggle')).toBeTruthy()
+    expect(wrapper.emitted('item:toggle')[0]).toEqual([true])
+  })
+
+  it('renders the slot content when the accordion is expanded', async () => {
+    await wrapper.setProps({ isOpen: true })
+    expect(wrapper.findComponent(DpTransitionExpand).isVisible()).toBe(true)
+  })
+
+  it('applies the correct font weight class based on the prop', async () => {
+    await wrapper.setProps({ fontWeight: 'normal' })
+    expect(wrapper.find('button').classes()).toContain('weight--normal')
+  })
+
+  it('applies the correct font size class based on the compressed prop', async () => {
+    await wrapper.setProps({ compressed: true })
+    expect(wrapper.find('span').classes()).toContain('font-size-medium')
+  })
+})
--- a/tests/DpAccordion.spec.js
+++ b/tests/DpAccordion.spec.js
@@ -59,11 +59,6 @@ describe('DpAccordion', () => {
     expect(wrapper.findComponent(DpTransitionExpand).isVisible()).toBe(true)
   })
 
-  it('applies the correct font weight class based on the prop', async () => {
-    await wrapper.setProps({ fontWeight: 'normal' })
-    expect(wrapper.find('button').classes()).toContain('weight--normal')
-  })
-
   it('displays a reduced font size if "compressed" is set to "true"', async () => {
     await wrapper.setProps({ compressed: true })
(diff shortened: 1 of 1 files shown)


tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+        dataCy: 'accordionToggle',
+        fontWeight: 'bold',
+        compressed: false,
+        isOpen: false,
+        title: 'Test Title'
+      }
+    })
+  })
+
+  it('renders component correctly', () => {
```
hwiem, line 20: I find this more accurate than 'it renders correctly'

```suggestion
  it('the component is mounted', () => {
```

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(wrapper.exists()).toBe(true)
+  })
+
+  it('renders with the correct title and default props', () => {
+    expect(wrapper.find('button').text()).toContain('Test Title')
+    expect(wrapper.find('button').classes()).toContain('weight--bold')
+    expect(wrapper.find('button').attributes('aria-expanded')).toBe('false')
+    expect(wrapper.find('button').attributes('data-cy')).toContain('accordionToggle')
+  })
```
hwiem, line 29: I think that an 'it' statement ideally should contain only one assertion (there may be exceptions), so that when you read the description, you instantly know what is being tested, and when the test fails, you instantly know what it is that is not working. Here, it might be the `title`, the `fontWeight`, the `aria-expanded` attribute or the `data-cy` attribute.
So I think we should split this 'it' statement into several:

```
it('displays the title passed in as a prop', () => {
    expect(wrapper.find('button').text()).toContain('Test Title')
})

it('displays the title as "bold" if "fontWeight" is set to "bold"', () => {
    expect(wrapper.find('button').classes()).toContain('weight--bold')
})

it('sets "aria-expanded" to false if "isOpen" is false', () => {
    expect(wrapper.find('button').attributes('aria-expanded')).toBe('false')
})

it('renders the given data-cy attribute', () => {
    expect(wrapper.find('button').attributes('data-cy')).toContain('accordionToggle')
})
```

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('toggles the accordion visibility when the button is clicked', async () => {
+    expect(wrapper.vm.isVisible).toBe(false)
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.vm.isVisible).toBe(true)
+    expect(wrapper.find('button').attributes('aria-expanded')).toBe('true')
```
hwiem, line 35: In my opinion, this should be in its own 'it' statement again:

```
it('toggles the value of the aria-expanded attribute when the button is clicked', async () => {
  expect(wrapper.find('button').attributes('aria-expanded')).toBe('false')
  await wrapper.find('button').trigger('click')
  expect(wrapper.find('button').attributes('aria-expanded')).toBe('true')
}
```

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('emits an event when the accordion is toggled', async () => {
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.emitted('item:toggle')).toBeTruthy()
+    expect(wrapper.emitted('item:toggle')[0]).toEqual([true])
+  })
+
+  it('renders the slot content when the accordion is expanded', async () => {
+    await wrapper.setProps({ isOpen: true })
+    expect(wrapper.findComponent(DpTransitionExpand).isVisible()).toBe(true)
```
hwiem, line 46: In my opinion, this does not test whether the slot content is rendered. We should pass in slot content and test that it does appear in the rendered component.

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.vm.isVisible).toBe(true)
+    expect(wrapper.find('button').attributes('aria-expanded')).toBe('true')
+  })
+
+  it('emits an event when the accordion is toggled', async () => {
+    await wrapper.find('button').trigger('click')
+    expect(wrapper.emitted('item:toggle')).toBeTruthy()
+    expect(wrapper.emitted('item:toggle')[0]).toEqual([true])
+  })
```
hwiem, line 42:
```suggestion
  it('emits the "item:toggle" event with correct payload when the accordion is toggled', async () => {
    await wrapper.find('button').trigger('click')
    const isVisible = wrapper.vm.isVisible
    expect(wrapper.emitted('item:toggle')).toBeTruthy()
    expect(wrapper.emitted('item:toggle')[0]).toEqual([isVisible])
  })
```

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('renders the slot content when the accordion is expanded', async () => {
+    await wrapper.setProps({ isOpen: true })
+    expect(wrapper.findComponent(DpTransitionExpand).isVisible()).toBe(true)
+  })
+
+  it('applies the correct font weight class based on the prop', async () => {
+    await wrapper.setProps({ fontWeight: 'normal' })
+    expect(wrapper.find('button').classes()).toContain('weight--normal')
+  })
```
hwiem, line 52: I think this duplicates the test from above:

```
it('displays the title as "bold" if "fontWeight" is set to "bold"', () => {
    expect(wrapper.find('button').classes()).toContain('weight--bold')
})
```

tests/DpAccordion.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+  it('applies the correct font size class based on the compressed prop', async () => {
+    await wrapper.setProps({ compressed: true })
+    expect(wrapper.find('span').classes()).toContain('font-size-medium')
+  })
+})
```
hwiem, line 58: I'm not very happy with checking for the existence of certain classes, because these can change frequently. Maybe it's something we should discuss in the team. I'd slightly adjust the description:

```suggestion
  it('displays a reduced font size if "compressed" is set to "true"', async () => {
    await wrapper.setProps({ compressed: true })
    expect(wrapper.find('span').classes()).toContain('font-size-medium')
  })
})
```

hwiem (changes requested):

tests/DpAccordion.spec.js
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/931/commits/39bda26351732a41df2e0d91e8bc265648cb09fa

tests/DpAccordion.spec.js
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/931/commits/e7c2753f269e9bd36634a93bc6fb0bacdb4f4ad6

PR #932, opened 2024-07-16
Make color tokens available to Tailwind config

spiess-demos:
This PR makes existing color tokens available to the Tailwind config, enabling color-related Tailwind utility classes. All classes are documented within the demosplan-ui storybook instance.

- As we do not want to expose palette colors for unguided use, only brand, data, and semantic tokens are made available
- Tailwind classes always start with the name of the property the style is applied to, so the semantic token source file has been duplicated with respect to the new structure (`color.ui-tailwind.json`)
- To automatically render documentation, two new react components, `ColorTable` and `ColorTableGrouped` have been added
- A new transformGroup is added to the StyleDictionary Build which is responsible for the tailwind color configs.
- The doc page for color has been updated to focus on available semantic colors and only mention the palette colors for reference

Note that both the structure of the json token source files and the logic which actually transforms those into SCSS and Tailwind config can be refactored to be better readable in the future. 

### Try it out

- Run `yarn install`, `yarn upgrade jest` and `yarn storybook`
- Browse to `?path=/docs/design-tokens-color--docs` and watch the colors
- Open a demosplan-ui component file and apply a Tailwind Utility class - the storybook page for that component should show the color


salisdemos (approved):
works like a charm

PR #937, opened 2024-07-17
chor (refs DPLAN-11885): add Jest Test for the DpBadge component.

ahmad-demos:
https://demoseurope.youtrack.cloud/issue/DPLAN-11885/Unit-Tests-schreiben

- add Jest Test for the DpBadge component.


tests/DpBadge.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    wrapper = shallowMount(DpBadge, {
+      props: {
+        color: 'default',
+        size: 'medium',
+        text: 'Test Badge'
+      }
+    })
+  })
+
+  it('renders the correct text', () => {
```
hwiem, line 17:
```suggestion
  it('renders the text passed in as prop', () => {
```

tests/DpBadge.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    expect(wrapper.text()).toBe('Test Badge')
+  })
+
+  it('applies the correct color class', () => {
+    expect(wrapper.classes()).toContain('color-text-default')
+    expect(wrapper.classes()).toContain('bg-color-light')
+  })
```
hwiem, line 24: Maybe we could create a test for each value that the color prop can have? And we could adjust the code like this:

```suggestion
it('applies the correct colors for color=default', () => {
  const defaultClasses = ['color-text-default', 'bg-color-light']
  defaultClasses.forEach(class => {
    expect(wrapper.classes()).toContain(class)
  })
})
```

Then we could remove the "renders with different color" test below

tests/DpBadge.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+  it('applies the correct size class', () => {
```
hwiem, line 26: I'd do the same as described above for the color prop, and then we can remove the "renders with different size" test below

hwiem (changes requested):

tests/DpBadge.spec.js
ahmad-demos: Done in https://github.com/demos-europe/demosplan-ui/pull/937/commits/8b31152b6f065921247b5950e1d63fac50b74af1

salisdemos (approved):

hwiem (approved):

PR #945, opened 2024-07-19
feat: add fill color token

spiess-demos:
As yellow-dark-1 feels to dirty to be used on larger areas but charts need color, not background prop (and we can's use bg-status-[changed|progress|complete] for this reason), another token group is introduced which duplicates
text-status-[changed|progress|complete]-icon but uses yellow-base instead of yellow-dark-1.


salisdemos (approved):
:heart:

PR #946, opened 2024-07-19
feat(refs PC-138): add icons to DpIcon

hwiem:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-11882/Icons-fur-PlanningCloud-Hub-in-demosplan-ui-integrieren


src/components/DpIcon/util/iconConfig.ts
```diff
@@ -0,0 +1,184 @@
...
+  'mail': 'landscape', // alias for envelope-simple
+}
+
+// EXPORTS
+
+export const iconComponents: Record<IconName, Component> = {
+  ...mappedIconNames,
+  ...mappedAliasedIconNames,
+  ...mappedIconAliases
+}
```
spiess-demos, line 172: I'm afraid i don't understand the logic how iconComponents are calculated from the objects defined above. Maybe you could give me a walktrough here?

src/components/DpIcon/DpIcon.vue
```diff
@@ -1,57 +1,90 @@
...
+   * Small: 16
+   * Medium: 20
+   * Large: 24
+   * Xlarge: 32
+   */
+  size: {
+    type: String as PropType<IconSize>,
+    required: false,
+    default: 'medium',
+    validator: (prop: IconSize) => ['small', 'medium', 'large', 'xlarge'].includes(prop)
```
spiess-demos, line 49: Could we drop the repeated values here by doing something like 
```suggestion
    validator: (prop: IconSize) => Object.keys(SIZES).includes(prop)
```

src/components/DpIcon/util/iconConfig.ts
hwiem: In each of the objects, the icon name is mapped to the icon vue component. I created several objects to separate icon names that are already identical to the ones used by phosphor-icons/vue from icon names that are aliases (= not identical to the names used by phosphor-icons/vue) and icon names that are not currently used, but should replace the aliases in the future. Then these 3 objects are merged into one, so all icon names are available for use.
Not sure if that answers your question? If it does, maybe we could find better names for the variables?

src/components/DpIcon/DpIcon.vue
hwiem: Thanks, done in f5ad4b30416df009ba709e10493f8e2b79352ac8

spiess-demos:
> icon names: I suggest to move to using the phosphor icon names instead of using aliases (e.g. 'search' as alias for 'magnifying-glass') and to deprecate the aliases. 

I disagree. While it may reduce complexity to remove this extra mapping layer, on the other hand the custom icon names also carry a semantic meaning: `search` clearly communicates in which context the icon is to be used, while `magnifying-glass` is not.

src/components/DpIcon/util/iconConfig.ts
```diff
@@ -0,0 +1,184 @@
...
+  'tag': PhTag,
+  'user': PhUser,
+  'users-three': PhUsersThree,
+  'warning': PhWarning,
+  'x': PhX,
+  'x-circle': PhXCircle
+}
+
+// icon names that are not used yet but should replace the aliases
+const mappedAliasedIconNames: Record<AliasedPhosphorIconName, Component> = {
```
spiess-demos, line 91: Why are you passing `AliasedPhosphorIconName` here, but in line 59 `PhosphorIconName`? I see no difference in the two objects, both consist of a string corresponding to a PhosphorIcon component.

src/components/DpIcon/util/iconConfig.ts
spiess-demos: Thanks for the clarification, but i still do not understand why you differentiate between mappedIconNames and mappedAliasedIconNames. I'd just put everything in one object and have the aliases in the second.

Also, as all of these objects actually contain components already, i'd suggest to remove the "names" part:

```suggestion
export const iconComponents: Record<IconName, Component> = {
  ...mappedIcons, // Contains both mappedIconNames + mappedAliasedIconNames
  ...mappedAliasedIcons // Renamed from mappedIconAliases
}
```

spiess-demos (changes requested):

src/components/DpIcon/util/iconConfig.ts
hwiem: The `AliasedPhosphorIconName` type contains the names that are aliases for phosphor icon names, the `PhosphorIconName` type contains names that are identical to phosphor icon names. My idea was to separate these so removing the aliases will be easier later.  I found it rather time consuming to understand which icon name corresponds to which phosphor icon, but maybe it's easier now that we map the names to phosphor icon components directly.

src/components/DpIcon/util/iconConfig.ts
hwiem: I separated them because I thought it would make things easier to understand, but maybe that's not really working :sweat_smile: Changed in 096c044e42096163c54465b37e559a52e20a4032

spiess-demos (approved):
lgtm!

hwiem:
> > icon names: I suggest to move to using the phosphor icon names instead of using aliases (e.g. 'search' as alias for 'magnifying-glass') and to deprecate the aliases.
> 
> I disagree. While it may reduce complexity to remove this extra mapping layer, on the other hand the custom icon names also carry a semantic meaning: `search` clearly communicates in which context the icon is to be used, while `magnifying-glass` is not.
> 
> When looking through the aliases, i'd agree that not all are needed. The `chevron-` icons could be removed, as well as `userSolid` which can be implemented with the `weight` prop.
> 
> If we really want to deprecate the aliases, we should possibly find another way of expressing which icon is to be used to communicate which context.

For the record: We decided to keep the aliases, as mapping them to the phosphor icon names is now easier (since we use phosphor-icons/vue). They need some cleaning up (removing some, maybe adding others), but this will be done at a later point.

PR #950, opened 2024-07-22
Refactor StyleDictionary script

spiess-demos:
This PR moves the logic to build tokens into /scripts/buildTokens.js and splits out helper functions into a separate utils/transformTokens.js file. It furthermore enhances the StyleDictionary build so that all config files are transformed directly to be consumable by Tailwind, not only the color files.

### How to test

- `yarn build:tokens` should produce...
  - the same scss files as before
  - the same tailwind config files, except that...
    - color.js now contains brand, palette, and generic semantics
    - 'boxShadow', 'breakpoints', 'rounded', 'space', 'zIndex' are transformed as files rather than on the fly
    - 'fontSize' is now also rendered as tw config file


scripts/buildTokens.js
```diff
@@ -0,0 +1,118 @@
...
+  transforms: StyleDictionary.transformGroup.web.concat(['name/web'])
+})
+
+/**
+ * Custom format that generates javascript modules ready to be used as Tailwind config files.
+ * The values use a nested set of css variables to enable theming on multiple levels.
+ */
+StyleDictionary.registerFormat({
+  name: 'tailwind/variables',
+  formatter: (formatterArguments) => transformTailwindTokens(formatterArguments, corePluginsColor, corePlugins)
```
salisdemos, line 56: You call the function `transformTailwindTokens` with 3 params. But as far as I see it only accepts 2. should it accept 3 Or should it not handle the corePlugins?

scripts/buildTokens.js
```diff
@@ -0,0 +1,118 @@
...
+  .sync(tokensPath)
+  .map(filePath => filePath
+    .replace('tokens/src/', '')
+    .replace('color/', '')
+    .replace('.json', ''))
+    .filter(filePath => !filePath.startsWith('_')) // Do not render tokens only used internally
+
+// Define custom output groups for Tailwind, rather than just grouping them by file path
+const corePlugins = ['fontSize', 'boxShadow', 'breakpoints', 'rounded', 'space', 'zIndex']
+const corePluginsColor = ['color', 'textColor', 'backgroundColor', 'borderColor']
```
salisdemos, line 22: By now I can't see the point why its necessary to divide corePlugins and corePluginsColor.

scripts/buildTokens.js
```diff
@@ -0,0 +1,118 @@
...
+  platforms: {
+    js: {
+      transformGroup: 'js',
+      buildPath: 'tokens/dist/js/',
+      files: files.map((filePath) => {
+        return {
+          destination: `${filePath}.js`,
+          format: 'javascript/module',
+          filter: (token) => {
+            return token.filePath.includes(filePath) && token.$status !== 'Deprecated'
```
salisdemos, line 80: I thought deprecated means its not recommended for usage and will be removed "soon", but still available. With this. Its not available at all if I am correct.

salisdemos (changes requested):
First of all: I looks really good.
Some Thinks at the first glance. But I love to get a walk through.

scripts/buildTokens.js
spiess-demos: thanks! i forgot to remove it. 35e3df5fdeab2ec81cd0aaf2fdccb3eacc623273

scripts/buildTokens.js
spiess-demos: This is correct, however the `'javascript/module'` format used in the `js` transformGroup produces js modules that are only used for the Storybook documentation.

For the Tailwind transform, i also have chosen to only output non-deprecated tokens (see `isDeprecated` in transformTokens.js). The "legacy tokens" are only available in Scss, as we're not gonna use them in Tailwind. At the moment, the `Deprecated` status is used exclusively for tokens which are renamed to match the Tailwind naming scheme, eg. in boxShadow.json, where the Scss naming scheme was `box-shadow-1`..`box-shadow-4` while Tailwind uses T-Shirt-Sizes. The deprecated tokens are simply aliases to the new names.

salisdemos (approved):
looks good. Tokens get generated.

PR #953, opened 2024-07-23
chore: Bump Yarn to 4.2.2

salisdemos:
- to stay in sync with -core, we should bump yarn to 4.2.2 too
- Add missing peer-dependencies which where pointed out by yarn. Now the output is clean
- bump node to 18.19.0 like in -core


package.json
```diff
@@ -2,7 +2,6 @@
   "name": "@demos-europe/demosplan-ui",
   "version": "0.3.21",
   "license": "MIT",
-  "private": false,
```
salisdemos, line 5: I don't know if we need that, but yarn kicks this line out automatically all the time.

package.json
```diff
@@ -65,24 +64,36 @@
...
     "@storybook/addon-styling-webpack": "^0.0.6",
+    "@storybook/blocks": "^7.0.27",
+    "@storybook/components": "^7.0.27",
+    "@storybook/core-events": "^7.0.27",
+    "@storybook/preview-api": "^7.0.27",
+    "@storybook/theming": "^7.0.27",
+    "@storybook/types": "^7.0.27",
     "@storybook/vue": "7.6.17",
     "@storybook/vue-webpack5": "7.6.17",
+    "@storybook/vue3": "^7.0.27",
```
hwiem, line 75: Why is this here?

package.json
```diff
@@ -105,6 +117,7 @@
     "ts-jest": "^29.1.0",
     "ts-loader": "^9.4.4",
     "typescript": "~5.5.3",
+    "vue": "^2.7.15",
```
hwiem, line 120: Should this be here? (It's also in peer dependencies)

spiess-demos (commented):
Looks good. maybe we can add `.yarn/install-state.gz` to gitignore?

package.json
salisdemos: Yes. Its necessary for some dependencies during the build process. peer dependency only explains, that the leading app has to require it too,

package.json
salisdemos: Actually I wonder too. But yarn pointed out, that was an unresolved dependency. Since storybook is sort of standalone, I can imagine, that storybook 7 uses vue3. Maybe we don't need @storybook/vue... let me find out....

spiess-demos: storybook 8 uses vue3, afaics.

salisdemos: > storybook 8 uses vue3, afaics.

For me it looks like SB 7.6.17 already runs with vue3. at least in some sort of compat mode.

spiess-demos (approved):
lgtm (code review only)

hwiem (approved):
lgtm

PR #955, opened 2024-07-24
fix(refs DPLAN-12137): Pass Data to tableHeader

salisdemos:


spiess-demos (approved):
lgtm (code review only).
maybe this deserves an entry in Changelog?

PR #971, opened 2024-07-31
Style DpInput, DpLabel, and Tooltip via Tailwindcss

spiess-demos:
Ticket: [DPLAN-11883](https://demoseurope.youtrack.cloud/issue/DPLAN-11883/demosplan-styles-durch-tailwind-ersetzen)

This PR adds styling via Tailwindcss to the DpInput and DpLabel component, as well as the Tooltip directive. To make this work, several things had to be done:

- **Font Sizes vs. Line height**: in demosplan-core, all font size utilities come with a unique line-height. Now, Tailwind is configured to do the same. A word on that: i made it so that the `sm` and `lg` sizes share a line-height of 1.4, while `base` goes with 1.5 because that is the best default for readable text of 16px size. All headings except h4 (which is the same as base font size) also go with 1.3 as it is implemented in demosplan-core. This mapping is done in the tokens file, but sadly StyleDictionary does not support composite tokens in v3 out of the box, so i rolled my own.
- **Other new tokens**: `rounded-input`, `border-input`, `text-input` and `text-input-disabled` join the tokens party to make theming inputs possible. `surface-dark` is the new old background for tooltips.
- **Invalid Styles**: As field validation is done from outside of DpInput, the respective styles are implemented globally in the tailwind.config file. 
- **Global text color**: To have a sensible default for elements that normally inherit color from html (like the label element, which is not colored separately in demosplan), the tailwind.config comes with a base color now. This is derived from the `textColor.default` token.
- lengthhint has been refactored to use tw color utilities.
- **Null default**: several props which defaulted to false now default to null because vue3 seems to handle attrs with false value other than vue2, resulting in them being rendered.
- **Tooltip overhaul**: When refactoring classes for the tooltip directive (which is used by DpLabel), i made some opinionated tweaks: slightly rounded corners, and slightly less padding.
- **Input Styling**: When working on the input styles, i realized that input fields were rounded like 3px in demosplan, while buttons were rounded 4px. Those values should be the same, so input fields have 4px border-radius now, too. Also, the focused state for input fields is a 2px border now, to better match the invalid state.

image


src/components/DpInput/DpInput.vue
```diff
@@ -37,248 +37,252 @@
...
+    cssClasses.push('bg-surface-light border-none cursor-default')
+  } else {
+    cssClasses.push('text-input bg-surface border border-input cursor-text')
+  }
+
+  if (props.hasIcon) {
+    cssClasses.push('pr-4')
+  }
+
+  // o-form__control-search is nowhere styled directly
```
hwiem, line 248: Can this be removed?

src/components/DpInput/DpInput.vue
```diff
@@ -37,248 +37,252 @@
...
+})
+
+const emit = defineEmits(['blur', 'enter', 'focus', 'input'])
+
+const { value, maxlength, minlength, width, size, label } = toRefs(props)
+const currentValue = ref(value.value)
+
+const classes = computed(() => {
+  let cssClasses: string[] = [
+    'px-1 py-0.5 max-w-full rounded-input',
```
hwiem, line 227: Why several strings?

src/components/DpInput/DpInput.vue
```diff
@@ -37,248 +37,252 @@
...
+    cssClasses.push('pr-4')
+  }
+
+  // o-form__control-search is nowhere styled directly
+  // cssClasses = props.type !== 'search' ? [...cssClasses, 'o-form__control-input'] : [...cssClasses, 'o-form__control-search']
+
+  return cssClasses.join(' ')
+})
+
+const containerClasses = computed(() => [
```
hwiem, line 254: nitpick: I don't find this very readable, I'd prefer it this way:

```
const classes = []

if (width.value !== 'auto' && size.value && size.value > 0) {
  classes.push(width.value)
}

classes.push(labelHint.value.length ? 'space-y-1' : 'space-y-0.5')

return classes
```
I know it's more verbose, but I understand it instantly.
But maybe it's something to discuss in the team at some point.

src/components/DpInput/DpInput.vue
```diff
@@ -37,248 +37,252 @@
...
+    default: null
+  },
 
-  components: {
-    DpLabel
+  /**
+   * Tell the browser if autocomplete is allowed or not. If enabled the browser is allowed
+   * to automatically complete the input. You can also provide a type of data which is expected.
+   */
+  autocomplete: {
```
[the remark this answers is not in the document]
spiess-demos: What would be the advantage in doing so? As of my understanding, browsers already apply heuristics to apply autocomplete behavior contextually when omitting the attr. Disabling it by default would also make it less visible to developers.

src/components/DpInput/DpInput.vue
```diff
@@ -37,248 +37,252 @@
...
+    type: String,
+    required: false,
+    default: ''
   },
 
-  data () {
-    return {
-      currentValue: this.value
-    }
+  dataCy: {
```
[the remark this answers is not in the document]
spiess-demos: This is correct. However cleaning this up across all props takes a little bit more time, as the value of some props is always expected to be string and this must be catched, then... so i'd like to defer this to a later point in time.

src/components/DpInput/DpInput.vue
spiess-demos: i did some mix with 6bb17e6f31e49af5e6365b9e5a7cc5dfd89fcebb

src/components/DpInput/DpInput.vue
spiess-demos: for better readability (several lines). 
changed to string literal in 6bb17e6f31e49af5e6365b9e5a7cc5dfd89fcebb

src/components/DpInput/DpInput.vue
spiess-demos: absolutely. 6bb17e6f31e49af5e6365b9e5a7cc5dfd89fcebb

src/components/DpInput/DpInput.vue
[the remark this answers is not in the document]
hwiem: It confused me a little that the default is an empty string, which results in autocomplete being 'on'; but you're right, we can keep it like this.

hwiem (approved):
lgtm!

src/components/DpInput/DpInput.vue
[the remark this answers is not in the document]
spiess-demos: in line 30:
`:autocomplete="autocomplete !== '' ? autocomplete : null"`
this means that the `autocomplete` attr is not rendered when the prop is not set.

PR #976, opened 2024-08-05
feat (refs AB#14276): Add new icons

gruenbergerdemos:
### Ticket
https://www.dev.diplanung.de/DefaultCollection/BOP-HH/_workitems/edit/14276

Data protection needs some updates.


spiess-demos (commented):
does'nt this deserve an entry in changelog?

gruenbergerdemos:
> does'nt this deserve an entry in changelog?

Yes, but it is still only a draft.

spiess-demos (approved):
lgtm!

PR #983, opened 2024-08-08
fix: restore installability of package

spiess-demos:
Due to several things, the yarn.lock and package.json were out of sync, so that vue-loader was not installed anymore by yarn 1 and the build failed consequently.

This change restores the state of the package.json before bumping Yarn to 4, and it additionally removes "@webpack-cli/generators" which seems to be a remnant of bootstrapping the project build.

Also, this change bumps several dependencies to their latest resolvable minor versions as i deleted yarn.lock before running `yarn install` to have a clean start, again.

Also, this change defines resolutions for `string-width` and `wrap-ansi` to fix the string-width-error, again. See [this comment for an explanation](https://github.com/yarnpkg/yarn/issues/8994#issuecomment-2207025146).


hwiem (commented):
When I run `yarn build` (after running `yarn`), I get the string-width error and need to run `yarn upgrade jest` to make it work. This results in an updated `yarn.lock`. Should that be commited?

spiess-demos:
> When I run `yarn build` (after running `yarn`), I get the string-width error and need to run `yarn upgrade jest` to make it work. This results in an updated `yarn.lock`. Should that be commited?

i got the string-width error, too. but sadly, after running `yarn upgrade jest` i got no changes in yarn.lock. idk how to deal with that. maybe you could commit your changes?

hwiem (approved):
lgtm + works, thanks :heart:

PR #995, opened 2024-08-20
fix(refs dplan-12305): Prevent Cursor from jumping to End of file

salisdemos:
**Ticket** 
[DPLAN-12305](https://demoseurope.youtrack.cloud/issue/DPLAN-12305/Munchen-Cursor-springt-nach-Korrektur-an-Ende-von-Stellungnahmetext-LHM-Ticket-INC0438291)

If there was an obscured Tag in the text, the cursor jumped to the End of the text on edit. This was caused by an unnecessarily often triggered method which checked transformed these obscured tags to a proper format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


ahmad-demos (approved):
lgtm, Test + codeReview!

PR #1012, opened 2024-09-06
Fix token build breakpoints 

spiess-demos:
This fix is also applied to the 0.3.x branch in #1011 

We cannot use Css variables within media queries. As we do not need to override breakpoints anyway, at least for now, the primitive px value is taken instead.

The change leads to the file `tokens/dist/tailwind/breakpoints.js` looking like this:

```js
module.exports = {
  "sm": "620px",
  "md": "904px",
  "lg": "1200px",
  "xl": "1400px"
};
```

instead of this:

```js
module.exports = {
  "sm": "var(--dp-breakpoints-sm, 620px)",
  "md": "var(--dp-breakpoints-md, 904px)",
  "lg": "var(--dp-breakpoints-lg, 1200px)",
  "xl": "var(--dp-breakpoints-xl, 1400px)"
};
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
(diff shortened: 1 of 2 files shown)


sakutademos (approved):
it works, lgtm!

PR #1031, opened 2024-10-02
chore: use default chunkSize value (Infinity) for tus uploads

muellerdemos:
### Description:
This PR removes the file upload max chunkSize from the DpUpload component and uses now the default value of the prop which resolves to `Infinity`. This change is for the demosplan vue3 compatible version.


hwiem (approved):
lgtm, thanks! 
same side note as in the other PR :)

muellerdemos:
@hwiem like in the other PR: Added a comment (and an explanatory commit message) here https://github.com/demos-europe/demosplan-ui/pull/1031/commits/d52d1733e556f623a38debeb08cd046b6c1a7350

PR #1032, opened 2024-10-02
Chore remove max chunk size on uploads vue2 version

muellerdemos:
### Description:
This PR removes the max chunkSize limitation for tus uploads in `DpUpload.vue`. This is for the vue2 version of demosplan-ui.
The same change for the `main` branch (aka vue3 compatible) can be found here: https://github.com/demos-europe/demosplan-ui/pull/1031


sakutademos (approved):
thx, code review, lgtm

hwiem (approved):
lgtm, thanks!
just a side note: I'd have explained the change in the commit message, so we (or others) can still understand it in the future. Maybe it would be helpful to add a comment to the prop that it should not be changed "unless you know what you're doing", possibly linking to https://github.com/tus/tus-js-client/blob/main/docs/api.md?

muellerdemos:
@hwiem thank you, this is a really helpful advise. I must admit I have a tendency to keep commit messages really short (like only having a short bulletpoint but no further explanation), which is something I should overthink from time to time since these can be very useful for other devs indeed.

Added a comment (and an explanatory commit message) here https://github.com/demos-europe/demosplan-ui/pull/1032/commits/d1cd06984001ddd5a9b4152b3abfe1f533be4011

PR #1033, opened 2024-10-02
fix: some adjustments after the change to yarn 4.2.2

sakutademos:
fix: move 'vue' from peerDep to devDep; add @tiptap/extension-text-style to fix dep. issue


salisdemos (changes requested):
Can You please add a changelog entry quickly :pray:

mussbach (approved):

PR #1037, opened 2024-10-07
fix(refs DPLAN-12574): fix upload files label position

muellerdemos:
### Description:
The label appears a little off on different occurrences hence some css adjustments were needed to go from:
Screenshot_2024-10-07_14-14-48
to:
Screenshot_2024-10-07_14-15-31

### Related PRs:
PR for vue3 version: 
https://github.com/demos-europe/demosplan-ui/pull/1036

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 - ([#1029](https://github.com/demos-europe/demosplan-ui/pull/1029)) add new utility: capitalizeFirstLetter ([@sakutademos](https://github.com/sakutademos)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


salisdemos (commented):
Thanks. 
I think this is a legit fix.
But in order to get somewhat closer to tailwind I would love to seen the opportunity grabbed to remove all the layout-classes and u-x-of-y in. I can't tell how urgent this is, so this is just a dream and a wish.

muellerdemos:
@salisdemos I think right now could be actually a good opportunity to tackle this right away since this component does not include a lot of styling at all and there ain't that many bugs left for the update sprints. Also I agree that it may be a very good idea to have only tailwind and not mix up custom styles in it (well if possible).

muellerdemos:
something about this addon is odd, after having a closer look at local and dev both versions look different and some translations do not work locally. I need to investigate this further hence this PR will be changed to a draft state for now (there are higher prio bugs on the board and this one has a low prio)

muellerdemos:
This bug has been solved already and this makes this PR reduandant

PR #1042, opened 2024-10-08
fix(refs DPLAN-12516): restore closeButton slot

hwiem:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-12516/Unnotiger-Leerraum-im-Stellungnahme-Fenster

It seems that at some point, we had a slot for the close button in the modal (it is used in the statement modal in demosplan), but somehow it got lost. It is hereby restored. :sparkles: 

### Tasks
- [x] Update changelog


salisdemos (approved):
Thanks. I just wonder why/when this vanished

PR #1043, opened 2024-10-09
fix(refs dplan-12602): Allow DpEditor to dismiss HTML from Word

salisdemos:
**Ticket** [DPLAN-12602](https://demoseurope.youtrack.cloud/issue/DPLAN-12602/BLP-STN-Formular-es-wird-nicht-der-ganze-Inhalt-bei-Copy-Paste-aus-Word-ubernommen)

Since its really complicated to fetch all edgecases from msOffice Word "html", there now is an option to "protect" the editor by denying the html-input from word.

Additionally I optimized the regex a bit to be not so greedy and to better find breaks like \n or \r

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+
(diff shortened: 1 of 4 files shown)


hwiem (approved):
lgtm, code review only

PR #1050, opened 2024-10-15
refac: checkbox: add additional type to selected prop

sakutademos:
Description: This PR includes an additional type for the prop `selected`, since it can be a number. (like `organisation.paperCopy` in the `PaperCopyPreferences.vue` component)

[PR in Core
](https://github.com/demos-europe/demosplan-core/pull/3810)

### DIFF
--- a/src/components/DpSelect/DpSelect.vue
+++ b/src/components/DpSelect/DpSelect.vue
@@ -127,7 +127,7 @@ export default {
     },
 
     selected: {
-      type: String,
+      type: [ String, Number ],
       required: false,
       default: ''
     }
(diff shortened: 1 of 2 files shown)


src/components/DpSelect/DpSelect.vue
```diff
@@ -127,7 +127,7 @@ export default {
     },
 
     selected: {
-      type: String,
+      type: [ String, Number ],
```
salisdemos, line 130:
```suggestion
      type: [String, Number],
```

salisdemos (approved):
just a nitpic, otherwise legit

PR #1052, opened 2024-10-16
fix(refs dplan-12315): Pass Item to dataTable instead of header-fields.

salisdemos:
coming from DpDataTableExtended, the whole row-item should be available in slots, not the headerfield informations.

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -484,15 +484,18 @@ export default {
   },
 
   watch: {
-    headerFields () {
-      if (this.isResizable) {
-        this.$nextTick(() => {
-          const firstRow = this.tableEl.getElementsByTagName('tr')[0]
-          const tableHeaderElements = firstRow ? firstRow.children : null
+    headerFields: {
+      handler () {
+        if (this.isResizable) {
+          this.$nextTick(() => {
+            const firstRow = this.tableEl.getElementsByTagName('tr')[0]
+            const tableHeaderElements = firstRow ? firstRow.children : null
 
-          this.setColsWidth(tableHeaderElements)
-        })
-      }
+            this.setColsWidth(tableHeaderElements)
+          })
+        }
+      },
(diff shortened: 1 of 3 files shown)


src/components/DpDataTable/DpDataTable.vue
```diff
@@ -484,15 +484,18 @@ export default {
...
+        if (this.isResizable) {
+          this.$nextTick(() => {
+            const firstRow = this.tableEl.getElementsByTagName('tr')[0]
+            const tableHeaderElements = firstRow ? firstRow.children : null
+
+            this.setColsWidth(tableHeaderElements)
+          })
+        }
+      },
+      deep: false // HeaderFields are always replaces as a whole and therefor deep watch is not necessary
```
spiess-demos, line 498: I think this is the default for watchers. See [Vue3 docs](https://vuejs.org/guide/essentials/watchers#deep-watchers).

src/components/DpDataTableExtended/DpDataTableExtended.vue
```diff
@@ -65,24 +65,12 @@
         </slot>
       </template>
       <template
-        v-for="el in filteredFields"
-        v-slot:[el.field]="element">
+        v-for="el in [...filteredFields, { field: 'expandedContent' }, { fields: 'flyout' }]"
```
spiess-demos, line 68: If i correctly get the idea, i think it should be: 
 
```suggestion
        v-for="el in [...filteredFields, { field: 'expandedContent' }, { field: 'flyout' }]"
```

spiess-demos (changes requested):

src/components/DpDataTable/DpDataTable.vue
salisdemos: true. I added it, to make clear, that I checked the spot and its fine and the warning won't point me to that line again and again

spiess-demos (approved):
lgtm (code review), assuming that this targets the right branch.

PR #1053, opened 2024-10-16
fix(refs dplan-12315): Configure Option to deny drag/dropping on rootlevel

salisdemos:
Configure Option to deny drag/dropping on rootlevel
(and one minor adjustment)


sakutademos (approved):
code review, lgtm

PR #1054, opened 2024-10-16
fix(refs dplan-12315): Make multiselect validation vue3 compatible

salisdemos:
our validator targeted the wrong element for vue-muliselect in vue3


sakutademos (approved):
code review, lgtm

PR #1068, opened 2024-11-01
feat(refs DPLAN-3217): allow disabling secondary button

hwiem:
**Ticket:** [DPLAN-3217](https://demoseurope.youtrack.cloud/issue/DPLAN-3217/ADO-Issue-15115-Integration-EWM-Part-2-Stellungnahmedetailansicht-aus-EWM-und-dplan-vereinen)

When `disabled` is set to true, not only the primary button, but also the secondary button should be disabled.


salisdemos (approved):
Thanks!

spiess-demos (approved):
lgtm!

PR #1074, opened 2024-11-12
feat (refs DPLAN-3217 AB#12115): Allow numbers as value for input

gruenbergerdemos:
### Ticket
[DPLAN-3217](https://demoseurope.youtrack.cloud/issue/DPLAN-3217)

- DpInput: Also allow numbers as values
- DpEditableList: Use DpButton instead of button and use new phosphor icons instead of fontawesome

### Related PR
https://github.com/demos-europe/demosplan-core/pull/3896


salisdemos (approved):
looks good. thank You

PR #1078, opened 2024-11-12
feat: Allow "reset" as dp-button type

salisdemos:
To use DpButton in standard Forms as Resete button, we have to allow the button-Type "reset".


gruenbergerdemos (approved):

gruenbergerdemos:
Do we need a changelog entry for that?

salisdemos:
> Do we need a changelog entry for that?

Yes. I was just writing it. You where to fast.

hwiem:
Saw this only now. The same change has been done in https://github.com/demos-europe/demosplan-ui/pull/1086 which was already merged, so we can close this.

PR #1081, opened 2024-11-14
feat(refs DPLAN-3217, #AB15115): make DpUpload available to end users

hwiem:
Sometimes we want to use `DpUpload` directly, without displaying the list of uploaded files (in `DpUploadFiles`); this is hereby made possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Changed
+- ([#1081](https://github.com/demos-europe/demosplan-ui/pull/1081)) DpUpload: Make component available for direct usage ([@hwiem](https://github.com/hwiem))
+
 ## v0.3.36 - 2024-11-14
 
 ### Added


src/components/DpUploadFiles/DpUploadFiles.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/vue'
-import DpUploadFiles from './'
+import { DpUploadFiles } from './index'
```
salisdemos, line 2: why that change? I mean, it works, but why do You think this change is an improvement?

src/components/DpUploadFiles/index.ts
```diff
@@ -1,2 +1,2 @@
-export { default } from './DpUploadFiles.vue'
-export * from './DpUploadFiles.vue'
+export { default as DpUploadFiles } from './DpUploadFiles.vue'
+export { default as DpUpload } from './DpUpload.vue'
```
salisdemos, line 2: do I get it right, and we can all refactor our exports to this to simplify things?

salisdemos (approved):

src/components/DpUploadFiles/DpUploadFiles.stories.tsx
hwiem: Ah, the './index' part is not needed, or what are you referring to?

salisdemos:
```
import DpUploadFiles from './'
```
should still work or not?

hwiem: Ok, I think the option to import like that gets lost if we import two components in the index file (I believe it's not related to removing `export * from './DpUploadFiles.vue'`). But the truth is that we never import like that
`import DpUploadFiles from './'`

src/components/DpUploadFiles/index.ts
hwiem: If I understand correctly, we can remove this kind of export statement
`export * from './DpUploadFiles.vue'`
from our index files, since it exports the named exports from the component (which I believe we don't have)

PR #1082, opened 2024-11-14
feat(refs DPLAN-3217, #AB15115): make DpUpload available to end users

hwiem:
Sometimes we want to use `DpUpload` directly, without displaying the list of uploaded files (in `DpUploadFiles`); this is hereby made possible.


spiess-demos (approved):
lgtm! 
in my personal opinion the naming of these components is a little off... however that's just first world problems i guess.

hwiem:
> lgtm! in my personal opinion the naming of these components is a little off... however that's just first world problems i guess.

I agree that the naming is not ideal. Do you have naming suggestions (for a future PR)?

PR #1086, opened 2024-11-18
fix(refs DPLAN-12823): Add 'reset' type to DpButton

hwiem:
**Ticket:** [DPLAN-12823](https://demoseurope.youtrack.cloud/issue/DPLAN-12823)

This PR adds 'reset' to the possible values for the `type` prop in DpButton. This allows us to use DpButton as a reset button in a form.


gruenbergerdemos (approved):
lgtm

PR #1093, opened 2024-11-20
fix(refs DPLAN-12886): fix overlapping flyout by adjusting css [VUE3]

muellerdemos:
### Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-12886/Abschnitte-Tabelle-Wenn-alle-Spalten-gewahlt-sind-Ellipsis-lauft-uber-den-Schritt-Text

### Description:
This PR adds some small css adjustments to the `dp-table-row` component of `dp-data-table` and aims to prevent the flyout trigger button to be overlapped by a column when re-adjusting the size of the column.

### Related PRs:
Vue2 PR: https://github.com/demos-europe/demosplan-ui/pull/1094

### DIFF
--- a/src/components/DpDataTable/DpTableRow.vue
+++ b/src/components/DpDataTable/DpTableRow.vue
@@ -33,31 +33,41 @@
     <td
       v-for="(field, idx) in fields"
       :key="`${field}:${idx}`"
-      :class="{ 'c-data-table__resizable': isTruncatable }"
+      :class="{ 'c-data-table__resizable': isTruncatable } + ' overflow-hidden'"
       :data-col-idx="`${idx}`">
       <div
-        :class="[
-          isTruncatable ?? 'break-words',
-          isTruncatable
-            ? wrapped
-              ? 'c-data-table__resizable--wrapped'
-              : 'c-data-table__resizable--truncated'
-            : ''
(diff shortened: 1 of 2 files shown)


src/components/DpDataTable/DpTableRow.vue
```diff
@@ -33,7 +33,7 @@
     <td
       v-for="(field, idx) in fields"
       :key="`${field}:${idx}`"
-      :class="{ 'c-data-table__resizable': isTruncatable }"
+      :class="{ 'c-data-table__resizable': isTruncatable } + ' overflow-hidden min-w-[50px]'"
```
hwiem, line 36: In the vue 2 PR, there is no `min-w-[50px]` here. Why do we need it here, but not there? Or is it missing there?

muellerdemos: ah indeed we do not need the `min-w-[50px]` class here, this was added here accidentially. Thanks for the catch, removed the class here https://github.com/demos-europe/demosplan-ui/pull/1093/commits/180af1425462aff2c7c7ea28caddddb9ae6dfd2f

hwiem (approved):
lgtm, code review only

PR #1096, opened 2024-11-22
Chore fix demosplan UI tests [vue3]

muellerdemos:
### Description:
This PR fixes the tests for demosplan-ui [vue3] but I needed to skip some tests for `DpInput` and `DpTreeListNode`.
Mainly because `DpTreeListNode` still contains some vue2 logic and instead of fixing the tests we should update this component and remove the `$root.$on ` occurrences.

To resolve the icon issues, we simply could add them as mocks to the __mocks__ folder with the folder structure and naming provided with this PR.

However the main goal of this PR is to have a clean test run so I think we can skip these tests for now (even though I do not like the idea) but nevertheless, we need a new release of demosplan-ui for several bugfixes for deployments hence I think we can do this for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Fixed
 
+- ([#1093](https://github.com/demos-europe/demosplan-ui/pull/1096)) Fix jest tests ([@muellerdemos](https://github.com/muellerdemos))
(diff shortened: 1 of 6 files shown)


hwiem (commented):
Since we don't need this for a release after all, why don't we keep the fixes and remove the skipping of tests? And thanks for the fixes :heart:

muellerdemos:
@hwiem I agree, since we do not need this right now as a releasable version, I reverted the skips added with this PR here: https://github.com/demos-europe/demosplan-ui/pull/1096/commits/4a47f2468b7508ba5d0dd185f5057cea56f6938d

PR #1101, opened 2024-11-28
fix(refs DPLAN-12460): addjust style to always display flyout properly

muellerdemos:
### Ticket:
https://demoseurope.youtrack.cloud/issue/DPLAN-12460/Bei-Abschnittsliste-mit-wenigen-Eintragen-sind-nach-Klick-auf-drei-Punkte-die-Optionen-abgeschnitten

### Description:
- the hidden and visible overflow styles were swapped by accident ... sorry, this should fix the issue now (hopefully but I double tested this now and this should do the trick finally)

### DIFF
--- a/src/components/DpDataTable/DpTableRow.vue
+++ b/src/components/DpDataTable/DpTableRow.vue
@@ -67,7 +67,7 @@
 
     <td
       v-if="hasFlyout"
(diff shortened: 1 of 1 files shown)


salisdemos (approved):
Looks good. Even though I am not sure if we need the overflow-hidden on the expandable col.

PR #1116, opened 2024-12-11
feat(refs DPLAN-12967): fix table

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12967/ADO-Issue-21834-Admin-Institutionsliste-Institutionen-konnen-verschlagwortet-werden

Some fixes for DpDataTable:
- Added a check to verify firstChild is an element node (nodeType === 1) before accessing its style property to prevent errors when firstChild is a text node (nodeType === 3).
- Remove overflow hidden from normal table rows, because with overflow hidden content in the table row eg selects are not shown correctly anymore. Overflow hidden was introduced to fix this [ticket](https://demoseurope.youtrack.cloud/issue/DPLAN-12886/Abschnitte-Tabelle-Wenn-alle-Spalten-gewahlt-sind-Ellipsis-lauft-uber-den-Schritt-Text), but imo it is not necessary, because the columns are resizable and the user can easily adjust the column to make more space for the menu. Without resizable columns enough space is automatically applied. 

Related PR
https://github.com/demos-europe/demosplan-core/pull/4115

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -687,7 +687,7 @@ export default {
       // Remove styles set by initialMaxWidth and initialWidth after copying rendered width into th styles
       if (this.isResizable) {
         Array.from(tableHeaderElements).forEach(th => {
-          if (th.firstChild) {
+          if (th.firstChild && th.firstChild.nodeType === 1) {
             th.firstChild.style.width = null
             th.firstChild.style.maxWidth = null
             th.firstChild.style.minWidth = null
--- a/src/components/DpDataTable/DpTableRow.vue
+++ b/src/components/DpDataTable/DpTableRow.vue
@@ -33,7 +33,7 @@
     <td
       v-for="(field, idx) in fields"
       :key="`${field}:${idx}`"
-      :class="{ 'c-data-table__resizable': isTruncatable } + ' overflow-hidden'"
+      :class="{ 'c-data-table__resizable': isTruncatable }"
       :data-col-idx="`${idx}`">
       <div
         v-if="isTruncatable"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 - ([#1105](https://github.com/demos-europe/demosplan-ui/pull/1105)) BREAKING: Minimum required node version: 20.18.1 ([@salisdemos](https://github.com/salisdemos))
 
+### Fixed
+
+([#1116](https://github.com/demos-europe/demosplan-ui/pull/1116)) DpDataTable: Remove overflow hidden from normal table rows and add a node type check for first children ([@gruenbergerdemos](https://github.com/gruenbergerdemos))
+
 
 ## v0.3.41 - 2024-11-28
 


src/components/DpDataTable/DpTableRow.vue
```diff
@@ -33,7 +33,7 @@
     <td
       v-for="(field, idx) in fields"
       :key="`${field}:${idx}`"
-      :class="{ 'c-data-table__resizable': isTruncatable } + ' overflow-hidden'"
+      :class="{ 'c-data-table__resizable': isTruncatable }"
```
salisdemos, line 36: did You check why it was added? I am not sure, if we simply can remove it without re-adding other issues.

src/components/DpDataTable/DpDataTable.vue
```diff
@@ -687,7 +687,7 @@ export default {
       // Remove styles set by initialMaxWidth and initialWidth after copying rendered width into th styles
       if (this.isResizable) {
         Array.from(tableHeaderElements).forEach(th => {
-          if (th.firstChild) {
+          if (th.firstChild && th.firstChild.nodeType === 1) {
```
salisdemos, line 690: Just out of curiosity: what else can nodeType be?

salisdemos (changes requested):

src/components/DpDataTable/DpTableRow.vue
gruenbergerdemos: Yes, see second point above in PR description

src/components/DpDataTable/DpDataTable.vue
gruenbergerdemos: Here is a [list of all node types](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType). Only node type `element` (1) can have the style attribute.

salisdemos: But in This Case I mean. When is it no Element?

gruenbergerdemos: I see, it can be a text node

src/components/DpDataTable/DpTableRow.vue
salisdemos: > Because the columns are resizable and the user can easily adjust the column to make more space for the menu

I would argue against that. When I want to access the Navigation quickly, I don't want to slide the cols "all the time".
Can we leave it like it was and try to find a overarching solution afterwards?

gruenbergerdemos: You don't need to do it all the time, only once. The width of the columns is saved. On the other hand, with overflow-hidden the user can't access the dropdown in selects if the the table row is not big enough and can't use the feature. I agree, that an overarching solution would be nice, but I could'nt find one so far and imo this is a much bigger problem than the orange bug that was fixed with this.

salisdemos: :thinking:

salisdemos: I would love to put an explaining comment somewhere, but I don't know where. I see Your point. I didn't think about interactive elements within the cell-content.

salisdemos (approved):

PR #1120, opened 2024-12-17
bug(refs DPLAN-13085): Add break words to table

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-13085/Institutionen-gruppieren-Toggle-fur-Spalte-geht-uber-den-Kategorie-Text

- Add break words to table columns to avoid overlapping content with resizable columns.
- Make header truncatable

### DIFF
--- a/src/components/DpDataTable/DpResizableColumn.vue
+++ b/src/components/DpDataTable/DpResizableColumn.vue
@@ -2,7 +2,7 @@
   <th
     v-tooltip="headerField.tooltip || headerField.label"
     ref="resizableColumn"
-    class="c-data-table__resizable"
+    class="c-data-table__resizable break-words"
     :class="{ 'u-pr-0' : isLast }"
     :data-col-field="headerField.field"
     :data-col-idx="idx">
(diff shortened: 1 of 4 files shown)


src/components/DpDataTable/DpResizableColumn.vue
```diff
@@ -2,7 +2,7 @@
   <th
     v-tooltip="headerField.tooltip || headerField.label"
     ref="resizableColumn"
-    class="c-data-table__resizable"
+    class="c-data-table__resizable break-words"
```
muellerdemos, line 5: nitpick: we may also use tailwind here I guess, and that would be like that:
`hyphens-auto`

https://tailwindcss.com/docs/hyphens#auto

muellerdemos (approved):
i think it is fine but I would have used tailwind classes here instead

src/components/DpDataTable/DpResizableColumn.vue
gruenbergerdemos: I'm not sure what you mean here. `break-words` is a tailwind class. It is extended in the override in tailwind.config.js with `hyphens: auto`: 
`        '.break-words': {
          'word-wrap': 'break-word',
          'word-break': 'break-word',
          'hyphens': 'auto'
        },`

muellerdemos (approved):
thanks for improving this

PR #1122, opened 2024-12-18
feat(refs dplan 12967): fix table main

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-12967/ADO-Issue-21834-Admin-Institutionsliste-Institutionen-konnen-verschlagwortet-werden

Some fixes for DpDataTable:
- Added a check to verify firstChild is an element node (nodeType === 1) before accessing its style property to prevent errors when firstChild is a text node (nodeType === 3).
- Remove overflow hidden from normal table rows, because with overflow hidden content in the table row eg selects are not shown correctly anymore. Overflow hidden was introduced to fix this [ticket](https://demoseurope.youtrack.cloud/issue/DPLAN-12886/Abschnitte-Tabelle-Wenn-alle-Spalten-gewahlt-sind-Ellipsis-lauft-uber-den-Schritt-Text), but imo it is not necessary, because the columns are resizable and the user can easily adjust the column to make more space for the menu. Without resizable columns enough space is automatically applied. 

Related PR (0.3.x)
https://github.com/demos-europe/demosplan-ui/pull/1116


src/components/DpDataTable/DpTableRow.vue
```diff
@@ -33,7 +33,7 @@
     <td
       v-for="(field, idx) in fields"
       :key="`${field}:${idx}`"
-      :class="[{ 'c-data-table__resizable': isTruncatable }, { 'break-words': isResizable }] + ' overflow-hidden'"
+      :class="[{ 'c-data-table__resizable': isTruncatable }, { 'break-words': isResizable }]"
```
salisdemos, line 36: shou'nd be the class `c-data-table__resizable--truncated` even more sufficient instad of `break-words`?

salisdemos (changes requested):

src/components/DpDataTable/DpTableRow.vue
gruenbergerdemos: It was recommended by UX to show the the whole content and not truncate it with `...` or something similar.

salisdemos (approved):

PR #1144, opened 2025-01-22
chore: Add drop space for DpTreeList

salisdemos:
Within DpTreeList: If you want to drop a draggable element in an empty category, its not possible since there is no space to drop it into. Therefor I added a spacer for that case.

### DIFF
--- a/src/components/DpTreeList/DpTreeListNode.vue
+++ b/src/components/DpTreeList/DpTreeListNode.vue
@@ -55,6 +55,7 @@
       :is="draggable ? 'dp-draggable' : 'div'"
       :drag-across-branches="options.dragAcrossBranches"
       class="list-style-none u-mb-0 u-1-of-1"
+      :class="[(children.length <= 0 && draggable) ? 'o-sortablelist__empty' : '']"
       data-cy="treeListChild"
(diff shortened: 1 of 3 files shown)


src/components/DpTreeList/DpTreeListNode.vue
```diff
@@ -94,6 +95,12 @@
             v-bind="scope" />
         </template>
       </dp-tree-list-node>
+      <li
+        v-if="isBranch && children.length <= 0 && hasDraggableChildren"
```
spiess-demos, line 99: why `&& hasDraggableChildren`?  i understand that this element serves as a drop zone if there are no leaves yet. However in that case, hasDraggableChildren would be false, i guess?

salisdemos: hasDraggableChildren indicates only if it could have draggable children. Even if there are no children yet, this is true. It "only" checks if draggable is true and the leaves are draggable:
```
 this.isBranch && (this.options.dragLeaves || this.options.dragBranches) && this.draggable
 ```

spiess-demos (approved):
lgtm (code review)

PR #1145, opened 2025-01-23
feat(refs DPLAN-12969,#AB21841): Add new styling to filter flyout

gruenbergerdemos:
Ticket https://demoseurope.youtrack.cloud/issue/DPLAN-12969

A prop `variant` is added to `DpFlyout`. It can be set to `'dark'` to set a darker background-color for the flyout trigger. The default value is `'light'`, so if the prop is not set, the flyout trigger will look as before.

### DIFF
--- a/src/components/DpFlyout/DpFlyout.vue
+++ b/src/components/DpFlyout/DpFlyout.vue
@@ -7,7 +7,7 @@
       'o-flyout--padded': padded,
       'is-expanded': isExpanded,
       'o-flyout--menu': hasMenu,
-      'bg-surface-medium py-0.5 mx-0.5 rounded-md': mode === 'dark'
+      'bg-surface-medium rounded-md': mode === 'dark'
     }"
     v-click-outside="close"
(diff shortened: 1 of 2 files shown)


src/components/DpFlyout/DpFlyout.vue
```diff
@@ -67,6 +68,13 @@ export default {
       default: true
     },
 
+    mode: {
```
spiess-demos, line 71: Nitpick: in many other cases we use the term "variant" for this kind of variation.

spiess-demos (commented):
How about changing the appearance in all instances instead of make a

src/components/DpFlyout/DpFlyout.vue
hwiem: Good point, changed in 7fe095ea9732a4f6afb296e69b807a2c1f46bb77

spiess-demos (approved):
lgtm (code review)

PR #1167, opened 2025-02-06
b-DPLAN-12869: add idDisabled prop to dl-label to change cursor style

riechedemos:
Ticket
DPLAN-12869

Description: Added isDisabled prop to DpLabel to change cursor from pointer to auto when input is disabled.

How to review/test
Login/ Click on user icon choose "Mein Profil" or Organization

PR: https://github.com/demos-europe/demosplan-project-diplanbau/pull/275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,15 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+
+- ([#1167](https://github.com/demos-europe/demosplan-ui/pull/1167)) DpLabel: add isDisabled props to change cursor style ([@riechedemos](https://github.com/riechedemos))
+
+
 ## v0.3.45 - 2025-01-30
 
 ### Added
 
-- ([#1167](https://github.com/demos-europe/demosplan-ui/pull/1167)) DpLabel: add isDisabled props to change cursor style ([@riechedemos](https://github.com/riechedemos))
 - ([#1158](https://github.com/demos-europe/demosplan-ui/pull/1158)) DpTreeList: remove padding for draggable leaves and keep it only for categories ([@sakutademos](https://github.com/sakutademos)
 - ([#1139](https://github.com/demos-europe/demosplan-ui/pull/1139)) DpFlyout: Add unit test ([@hwiem](https://github.com/hwiem))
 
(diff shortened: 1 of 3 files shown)


src/components/DpLabel/DpLabel.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <label
     :class="prefixClass(['o-form__label flex', bold ? 'weight--bold' : 'weight--normal', hints.length > 0 ? 'has-hint' : '', hide ? 'sr-only' : ''])"
```
hwiem, line 3: We try to avoid using the `style` attribute, it's difficult to maintain. Instead of setting the style directly, we can use tailwind classes: 

```suggestion
    :class="prefixClass(['o-form__label flex', bold ? 'weight--bold' : 'weight--normal', hints.length > 0 ? 'has-hint' : '', hide ? 'sr-only' : '', isDisabled ? 'cursor-default' : 'cursor-pointer'])"
```

I'd use `pointer: default` here - you can't interact with the input, so changing the cursor may be confusing for users. (The cursor for the disabled input itself should also be adjusted to a default cursor in my opinion, so it's consistent)

hwiem (changes requested):
Thanks for fixing this :heart: I have some minor change requests

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Added
 
+- ([#1167](https://github.com/demos-europe/demosplan-ui/pull/1167)) DpLabel: add isDisabled props to change cursor style ([@riechedemos](https://github.com/riechedemos))
```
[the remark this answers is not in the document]
riechedemos: thank you! I changed it.

src/components/DpLabel/DpLabel.vue
riechedemos: thanks, fixed it.

src/components/DpLabel/DpLabel.vue
```diff
@@ -96,7 +103,7 @@ export default {
 
     labelFor () {
       return this.for
-    }
+    },
```
hwiem, line 106: this is quite the nitpick, but I think eslint does not approve of trailing commas :sweat_smile:

hwiem (approved):
Thanks, code lgtm! just 1 tiny nitpick

src/components/DpLabel/DpLabel.vue
riechedemos: done :) f46b5ccd7023896bc95001348992217755320178

PR #1172, opened 2025-02-11
b-DPLAN-14659: add keys to navigate to preventEditing.js

riechedemos:
DPLAN-14659:
Description: added keys to navigate with keyboard to custom tiptap extention preventEditing.js 
  
Testing:  Verfahren / Abwägungstabelle/ tab to 2. checkbox and check it/ try shift + arrow to mark text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


spiess-demos (approved):
Nice! Lgtm (code review only)

spiess-demos (changes requested):
On a second thought: could you add the change to the [CHANGELOG](https://github.com/demos-europe/demosplan-ui/blob/main/CHANGELOG.md)? you could chose between either "Added" or "Fixed" but i'd tend to chose the latter. Also see https://common-changelog.org/

PR #1183, opened 2025-02-19
bug(refs DPLAN-14948): Add two disabled props for button row

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-14948/Aktuelles-Textbaustein-einfugen-moglich-auf-Einfugen-zu-klicken-obwohl-keine-Textbaustein-hinzugefugt-ist

It should be possible to disable only the primary button, because often times it makes sense that the secondary button is still active.

No uses of `DpButtonRow` with disabled prop were found in demosplan-ui.

### DIFF
--- a/src/components/DpButtonRow/DpButtonRow.vue
+++ b/src/components/DpButtonRow/DpButtonRow.vue
@@ -6,7 +6,7 @@
       v-if="primary"
       :busy="busy"
       :data-cy="`${dataCy}:saveButton`"
-      :disabled="disabledPrimary"
+      :disabled="isDisabledPrimary"
       :text="primaryText"
       :variant="variant"
       @click.prevent="$emit('primary-action')" />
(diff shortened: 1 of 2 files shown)


src/components/DpButtonRow/DpButtonRow.vue
```diff
@@ -6,15 +6,15 @@
       v-if="primary"
       :busy="busy"
       :data-cy="`${dataCy}:saveButton`"
-      :disabled="disabled"
+      :disabled="disabledPrimary"
```
hwiem, line 9: This is a nice improvement! I'm wondering though if we want to keep the option to disable both buttons easily by setting `disabled` to `true`, and if we want to disable just one button we could pass `disabled: { primary: true }` or something like that? Maybe it adds unnecessary complexity. What do you think? Do you know what the most common use case is?
Also, if we keep `disabledPrimary` and `disabledSecondary`, I think we need to adjust usages in demosplan?

gruenbergerdemos: I decided to go with your approach. This way we also don't have to worry about breaking things eg. in addons. I also added the option to also pass the secondary button. It probably is not necessary, but I think it does not hurt. 
Done in [6ddeba1](https://github.com/demos-europe/demosplan-ui/pull/1183/commits/6ddeba1f334f03589c23efa9e0d2de3a2507ef43)

PR #1187, opened 2025-02-19
feat(refs DPLAN-14634): add ariaLabel prop to DpFlyout

hwiem:
**Ticket:** [DPLAN-14634](https://demoseurope.youtrack.cloud/issue/DPLAN-14634/ADO-Issue-22463-Admin-Institutionsliste-Filteroptionen-konnen-nach-Schlagwortkategorien-angepasst-werden)

Adds an `ariaLabel` prop to DpFlyout that can be set if only an icon is used as the flyout trigger.


spiess-demos (approved):
lgtm (code review).

PR #1192, opened 2025-02-20
bug(refs DPLAN-14980): Use dynamic time pattern

gruenbergerdemos:
[Ticket](https://demoseurope.youtrack.cloud/issue/DPLAN-14980/Verfahrensschritt-automatisch-umstellen-Zeitangabe-undefined)

- Build the time pattern from available hours and minutes to avoid user input that is not allowed.
- Only update time from manual input if it is a valid pattern.


salisdemos (approved):
Thanks. Looks good.

PR #1194, opened 2025-02-21
refac(DPLAN-15036): expand the DpTreeListNode only when children exist

sakutademos:
**Ticket:** https://demoseurope.youtrack.cloud/agiles/141-245/current?issue=DPLAN-15036

**Description:** This PR improves UX: when the treeListNode is empty, it should not be possible to expand it.

### DIFF
--- a/src/components/DpTreeList/DpTreeListNode.vue
+++ b/src/components/DpTreeList/DpTreeListNode.vue
@@ -96,12 +96,6 @@
             v-bind="scope" />
         </template>
       </dp-tree-list-node>
-      <li
-        v-if="isBranch && !hasChildren && hasDraggableChildren"
-        v-show="isExpanded"
-        class="ml-2 mr-4 mt-2">
-        <div class="o-sortablelist__spacer relative" />
-      </li>
     </component>
   </li>
 </template>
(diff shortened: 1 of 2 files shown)


src/components/DpTreeList/DpTreeListNode.vue
```diff
@@ -223,16 +223,20 @@ export default {
...
+    hasChildren () {
+      return this.children.length > 0
+    },
+
     hasDraggableChildren () {
       return this.isBranch && (this.options.dragLeaves || this.options.dragBranches) && this.draggable
     },
 
     hasToggle () {
-      return this.isBranch && (this.children.length > 0 || this.hasDraggableChildren)
```
salisdemos, line 231: I think I rember that I added the `hasDraggableChildren` Flag to ensure that empty branches can be opended if the list is draggable but the branch is empty to allow dropping elements there. I am not sure if there something changed in the meanwhile, but we have to ensure, that we still can drop elements into empty branches

sakutademos: I understand what you mean and have tested it: we can drop elements into the closed branch as well, it does not need to be opened for this functionality. I am also not sure if we need this part: [[commit link]](https://github.com/demos-europe/demosplan-ui/pull/1194/commits/3d6e5fc83811a79df0e4998447c80197565a790f), as it works without it and an extra spacer is not needed there.

tested it: added an element from another branch to an empty, closed branch → Success (the branch can now be expanded)

salisdemos (approved):
Thanks!!!

PR #1202, opened 2025-03-03
fix: Catch Full-width in dpFormRow

salisdemos:
width tailwind we got a new full-width-class wich has to be catched


hwiem (approved):
lgtm, thanks!

PR #1204, opened 2025-03-04
fix(refs DPLAN-15153): fix imports for mention extension [vue 3]

hwiem:
Ticket: [DPLAN-15153](https://demoseurope.youtrack.cloud/issue/DPLAN-15153/Benutzer-Menu-Verwalten-und-Plattformtools-Links-funktionieren-nicht-auf-Verfahrenstyp-bearbeiten-Seite)

The workaround for the mention extension import is no longer needed and it is now imported from the correct file in DpEditor.vue.


salisdemos (approved):
Thanks for updating

PR #1207, opened 2025-03-06
fix(refs DPLAN-15152): Make mention extention load

salisdemos:
**Ticket**
DPLAN-15152

It turned out, that cfe12c41fa32ee0136714f4698e61c5b6e36df39 did not fix the issue. We still need the workaround but from the correct path


gruenbergerdemos (commented):
Is the other solution in another PR or repo? I'm a little confused about the changelog entry, but no other changes.

salisdemos:
> Is the other solution in another PR or repo? I'm a little confused about the changelog entry, but no other changes.

Sorry. I have to check what went wrong while creating the PR. This is not what it suppose to be

PR #1208, opened 2025-03-06
fix(refs DPLAN-15186): Remove useless wrapping div

salisdemos:
with that the selector to find the element to be validated is working again and the multiselect is validatable again

It looks larger than it is. I only removed the wrapping div.
Ticket: DPLAN-15186


src/components/DpEditableList/DpEditableList.vue
```diff
@@ -24,7 +24,7 @@
             variant="subtle"
             @click.prevent="showUpdateForm(index)" />
           <dp-button
-            class="-ml-1.5"
+            class="ml-0.5"
```
salisdemos, line 27: sorry. this was supposed to be a separate PR

src/components/DpMultiselect/DpMultiselect.vue
```diff
@@ -1,94 +1,92 @@
...
-                type="button"
-                v-text="translations.deselectAll"
-                @click="$emit('unselect-all')">
-            </button>
-          </div>
-        </slot>
-      </template>
-    </vue-multiselect>
-  </div>
+  <vue-multiselect
```
hwiem, line 2: Is it safe to remove the div? (In the commit message of the commit adding the div it said, "this has to be done to make vue3 work")

hwiem (approved):
Lgtm, code review only

PR #1209, opened 2025-03-06
fix: Escape multiline break chars

salisdemos:
Using multilines in Js causes \n in the string. They appeard in the classlist, where they dont belong. So we had to escape them


gruenbergerdemos (approved):
lgtm

PR #1215, opened 2025-03-14
chore(refs DPLAN-12921): Reintroduce clickOutside

salisdemos:
the old deplendency which handled click-outside events, was not ported to vue3 so we had to find another dependency.
the package @vueuse/components provides a directive quite similar to what we used before. But to get this running, I had to Bump vue to the latest Version.

Now the three spots  splitbutton, flyout and datepicker should be closable by clicking outside the element again.


spiess-demos (approved):
When refactoring to the composition api, we could as well use the composable https://vueuse.org/core/onClickOutside/ 

however this looks legit for the time being.

PR #1216, opened 2025-03-18
bug(refs DPLAN-15145): Add prop for rounded input

gruenbergerdemos:
Ticket: https://demoseurope.youtrack.cloud/issue/DPLAN-15145/Suche-Button-ist-eckig-wenn-Suche-Input-Feld-abgerundet-ist

Add prop for rounded input to decide if the whole input should be rounded or only left or right. Sometimes we want to add an icon eg for searching at the end of the input and with the whole input rounded it looks strange.


hwiem (approved):
Nice!

PR #1224, opened 2025-03-27
b(refs: DPLAN-15161) add hasContent props for DpTreeListToggle

riechedemos:
## Ticket
DPLAN-15161

Add conditional rendering for the DpTreeListToggle component with a new hasContent prop and a computed property for evaluation.

Related PR: https://github.com/demos-europe/demosplan-core/pull/4506

### DIFF
--- a/src/components/DpTreeList/DpTreeList.vue
+++ b/src/components/DpTreeList/DpTreeList.vue
@@ -18,7 +18,7 @@
           <slot name="header" />
         </div>
         <dp-tree-list-toggle
-          v-if="hasContent"
+          v-if="treeData.length > 0"
           class="color--grey"
           data-cy="treeListNodeToggle"
           @input="toggleAll"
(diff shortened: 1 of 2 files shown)


src/components/DpTreeList/DpTreeList.vue
```diff
@@ -18,6 +18,7 @@
           <slot name="header" />
         </div>
         <dp-tree-list-toggle
+          v-if="hasContent"
```
salisdemos, line 21: shouldn't this be 
```suggestion
          v-if="treeData.length > 0"
```
? 
Or are there circumstances where we want to handle that independent?

salisdemos (changes requested):

src/components/DpTreeList/DpTreeList.vue
riechedemos: No, I didn't realize I could use this, Thx I changed it. 666140d25b132afd337e96297bb4348e81c521b2

salisdemos (approved):

salisdemos:
I think https://github.com/demos-europe/demosplan-core/pull/4506 is then obsolet and can be closed?

PR #1229, opened 2025-04-02
build: set up eslint

hwiem:
Sets up eslint for demosplan-ui and autofixes a large number of errors.

The rest should be fixed in separate PRs.


elmasdemos (approved):
lgtm, nice work!! thanks for the effort! <3

PR #1230, opened 2025-04-02
test(refs DPLAN-15256, #AB9149): fix unit tests

hwiem:
Fixes the failing unit tests - but some warnings remain. I suggest to fix these in a separate PR.


spiess-demos (approved):
tests are fixed, thanks for the effort!

PR #1234, opened 2025-04-04
b(refs: DPLAN-15410) add padding to align trigger flyout btn

riechedemos:
Ticket
DPLAN-15410

Add padding to align trigger flyout btn and add some spacing to reset btn


salisdemos (approved):
Thanks a lot

PR #1236, opened 2025-04-07
fix(DPLAN-DPLAN-15342): update currentValue when the user inputs text

sakutademos:
**Ticket:** https://demoseurope.youtrack.cloud/agiles/141-63/current?settings&tab=allgemeines&issue=DPLAN-15342

**Description:** This PR fixes an issue where the `DpResettableInput` was only updated after the second input. It now assigns `this.currentValue = val `when the user inputs text.

 There is also a warning about deprecated v-model usage, which I tried to fix, but it does not work as expected. Here is a [thread about the issue](https://demosdeutschland.slack.com/archives/C5BA8ABB2/p1744028862868659). it can be fixed in a separate PR.

### DIFF
--- a/src/components/DpResettableInput/DpResettableInput.vue
+++ b/src/components/DpResettableInput/DpResettableInput.vue
@@ -8,7 +8,7 @@
       :required="required"
       :rounded="rounded"
       @blur="$emit('blur', currentValue)"
-      @input="$emit('input', currentValue)"
+      @input="onInput"
       @enter="$emit('enter', currentValue)"
       @focus="$emit('focus')"
       :pattern="pattern"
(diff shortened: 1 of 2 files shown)


src/components/DpResettableInput/DpResettableInput.vue
```diff
@@ -128,13 +128,12 @@ export default {
     }
   },
 
-  watch: {
-    value: function () {
-      this.currentValue = this.value
-    }
-  },
-
```
gruenbergerdemos, line 136: We still need this watcher, because resetting is handled in the parent via the value.

gruenbergerdemos (changes requested):
I was working on the same bug ^^ Had the same solution, but we still need the watcher.

gruenbergerdemos (approved):
Thanks

PR #1241, opened 2025-04-11
Feat DPLAN-15466 create dp radio buttongroup

riechedemos:
create dp radio buttongroup


src/components/DpRadioButtonGroup/DpRadioButtonGroup.vue
```diff
@@ -0,0 +1,67 @@
...
+  }
+})
+
+const emit = defineEmits(['update'])
+
+const selected = ref(props.selectedOption)
+
+const updateSelection = (option) => {
+  selected.value = option.id
+  emit('update', option)
```
salisdemos, line 61: I am not sure, but I was thinking if we want to have something like an "unselect" if the radiogroup is not required.

src/components/DpRadioButtonGroup/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './DpRadioButtonGroup.vue'
```
spiess-demos, line 1: Do we really want to name it `DpRadioButtonGroup` when we have `DpRadio`, `DpCheckbox`, and `DpCheckboxGroup`?

PR #1242, opened 2025-04-11
fix(refs DPLAN-15338, #AB26283): set headerCellCount on mounted

hwiem:
**Ticket:** [DPLAN-15338](https://demoseurope.youtrack.cloud/issue/DPLAN-15338/ADO-Issue-26283-Hinzufugen-konfigurierbare-Felder-auf-Abschnittsebene-Part-1-Anlegen)

If we set `headerCellCount` on `beforeUpdate`, the `headerFields` length is still 0, and the `colCount` is not calculated correctly if the table is empty

**Before:**
table_before

**After:**
table_after

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -715,6 +715,8 @@ export default {
     }
 
     this.forceElementSelections(this.shouldBeSelectedItems)
+
+    this.headerCellCount = this.headerFields.length
   }
 }
 </script>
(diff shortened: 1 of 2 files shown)


src/components/DpDataTable/DpDataTable.vue
```diff
@@ -715,6 +715,8 @@ export default {
     }
 
     this.forceElementSelections(this.shouldBeSelectedItems)
+
+    this.headerCellCount = this.headerFields.length
```
salisdemos, line 719: since this is legit, I wonder why headerCellCount is no computed...

salisdemos (approved):

src/components/DpDataTable/DpDataTable.vue
hwiem: With DpDataTable, I'm always afraid of unintentionally breaking things, so I don't feel very adventurous here.. :see_no_evil:

salisdemos: true

PR #1243, opened 2025-04-11
feat: emit blur and focus events

hwiem:
To be able to handle `blur`, `focus` and `mousedown` events from DpButton, we need to emit them first.


sakutademos (approved):
code review, lgtm :)

PR #1246, opened 2025-04-15
b(refs: DPLAN-15440) Adjusted the headingLevel function 

riechedemos:
DPLAN-15440

Adjusted the headingLevel function to expect an object with a level property.


salisdemos (approved):
looks good. Thanks

PR #1248, opened 2025-04-17
chore: Update a bunch of dependencies

salisdemos:
to avoid conflicts in the yarn.lock file when updating them one by one, I simply batched them

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@tiptap/pm": "^3.0.0",
     "@tiptap/suggestion": "^3.0.0",
     "@tiptap/vue-3": "^3.0.0",
-    "@uppy/core": "^4.3.1",
+    "@uppy/core": "^4.4.4",
     "@uppy/drag-drop": "^4.1.2",
     "@uppy/progress-bar": "^4.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -32,9 +32,9 @@
     "@tiptap/suggestion": "^3.0.0",
     "@tiptap/vue-3": "^3.0.0",
     "@uppy/core": "^4.3.1",
-    "@uppy/drag-drop": "^4.0.5",
-    "@uppy/progress-bar": "^4.1.0",
-    "@uppy/tus": "^3.0.1",
+    "@uppy/drag-drop": "^4.1.2",
```
hwiem, line 35: Could we also upgrade `@uppy/core` to v4.4.4? (that is what `@uppy/drag-drop` and other dependencies request)

hwiem (commented):
Thank you :heart: Just a suggestion

package.json
salisdemos: sure f4395524fa4edb4776ac5d5dcd00268c098f4f6d

hwiem (approved):
Nice!

PR #1256, opened 2025-04-29
Move documentation content from dplanDocs to Storybook

spiess-demos:
As we're centralizing org based docs while further pushing Storybook as the single source of truth for demosplan-ui, these contents are split into those 2 new places: Storybook and Youtrack.


salisdemos (approved):
Thanks!

PR #1257, opened 2025-04-29
feat: use yarn4 in dev container

spiess-demos:
As we've updated to yarn 4, it needs to be available also in the Docker container for local development.


hwiem (approved):
Lgtm + works, thank you :heart:

PR #1262, opened 2025-05-05
fix(DPLAN-15557): DpUploadFiles: add new prop 

sakutademos:
**Description:** After migrating to Vue3, we now pass `demosplanUi` into our addons via props and render them using `component` tags where `this.$refs` no longer works on those `components`, so trigger the method `clearFilesList` using the props instead.

PR in Addon: https://github.com/demos-europe/demosplan-addon-demospipes/pull/300


salisdemos (approved):
legit

PR #1265, opened 2025-05-07
feat: remove DpTooltipIcon

spiess-demos:
This component was already deprecated since v0.1.13 and is now removed.


salisdemos:
> This component was already deprecated since v0.1.13 and is now removed.

Its still in use in newsAdminList.js

spiess-demos:
> > This component was already deprecated since v0.1.13 and is now removed.
> 
> Its still in use in newsAdminList.js

it is imported there, but it is not used. however there is a [PR that removes this unused import](https://github.com/demos-europe/demosplan-core/pull/4637).

salisdemos (approved):
Thanks.

PR #1266, opened 2025-05-07
fix(refs: DPLAN-15411): add space between numbers and text in the pagination

rafelddemos:
[DPLAN-15411](url)

Changes display 'inline-block' to 'flexbox' on the parent 'label'-element to add spacing in the pagination on the 'Stellungnahmen'-page.


salisdemos (commented):
looks legit. Thank You

salisdemos (approved):

PR #1268, opened 2025-05-07
refactor(refs DPLAN-15262): replace vue-omnibox

hwiem:
Ticket: [DPLAN-15262](https://demoseurope.youtrack.cloud/issue/DPLAN-15262/Suchfeld-auf-der-Startseite-ist-kaputt)


salisdemos:
@huellnerdemos may You have a look?

salisdemos:
@huellnerdemos Test me again

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -1,114 +1,396 @@
...
+   */
+  routeGenerator: {
+    type: Function,
+    required: true
+  },
+
+  /**
+   * Current value of the input (v-model)
+   */
+  modelValue: {
```
sakutademos, line 143: I think the core should also be updated to use the model-value prop instead of v-model, because `currentAutocompleteSearch` remains empty. So that clicking the save button produces no results. Is there a PR for this?

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -1,114 +1,396 @@
...
+        :style="boxHeightStyle"
+        class="w-auto max-w-full py-[2px] text-gray-300 overflow-hidden whitespace-pre flex items-center"
+        aria-hidden="true">
+        {{ completion }}
+      </div>
+    </div>
+
+    <!-- Options list -->
+    <div class="relative">
+      <ol
```
sakutademos, line 54: sugesstions list should have a higher z-index to prevent this issue or the list below should have a lower index
image

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -1,114 +1,396 @@
...
-      v-for="(_, slotName) in $slots"
-      v-slot:[slotName]="slotData">
-      <slot
-        v-bind="slotData"
-        :name="slotName" />
-    </template>
-  </vue-omnibox>
+  <div class="relative">
+    <!-- Search container -->
+    <div
```
sakutademos, line 4: i am curious, why is a div used here instead of a native input? A input already provides styling and built-in behaviors like text selection, clipboard and so on. For example i can not highlight or copy the text in the search field, which can be problematic...

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -1,114 +1,396 @@
...
+  }
+}
+
+// Setup mutation observer to track contenteditable changes
+onMounted(() => {
+  if (!input.value) {
+    return
+  }
+
+  const observer = new MutationObserver((mutations) => {
```
sakutademos, line 376: should the MutationObserver be cleaned up in the onUnmounted hook?

src/components/DpAutocomplete/DpAutocomplete.vue
spiess-demos: `z-flyout` would do, i guess.

src/components/DpAutocomplete/DpAutocomplete.vue
salisdemos: can't reprocude, what You mean. For me it works when I test it in public_index or public_detail. Even without modelValue.
But You are right. I should adjust it in core too..

src/components/DpAutocomplete/DpAutocomplete.vue
salisdemos: It was like that in the old vue-omnibox. There where several issues, trying it with an input-element, so I decided to stick with the already working approach. Even though it comes with drawbacks. I love to find out whats going on, so we can use the native element, but not now :-/

src/components/DpAutocomplete/DpAutocomplete.vue
salisdemos: good point. Thanks for spotting
738d4666e9f0683579e3cb5d79d897f8bb55baae

sakutademos (approved):
thx!

PR #1269, opened 2025-05-07
Tailwind classes for some more components

spiess-demos:
This PR migrates the following components to using Tailwind classes:

- DpAccordion
- DpBadge
- DpBulkEditHeader
- DpCard
- DpFlyout

Also, it adds a Welcome page to Storybook, which (alongside other information) contains a list which tracks Tailwind migration process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 - ([#1269](https://github.com/demos-europe/demosplan-ui/pull/1269) BREAKING: Update DpAccordion, DpBadge, DpBulkEditHeader, DpCard, DpFlyout to use Tailwind classes ([@spiess-demos](https://github.com/spiess-demos))
 
+
+## v0.4.18 - 2025-06-18
+
+### Fixed
+- ([#1293](https://github.com/demos-europe/demosplan-ui/pull/1293)) DpSlidingPagination: add a check to emit the 'page-change' event when the page number does not exceed the total page count ([@sakutademos](https://github.com/sakutademos)
+- ([#1278](https://github.com/demos-europe/demosplan-ui/pull/1287)) DpTreeList: Implement node selection without mutating vuex store state ([@hwiem](https://github.com/hwiem))
+
+
 ## v0.5.1 - 2025-05-19
 
 ### Changed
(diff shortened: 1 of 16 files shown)


salisdemos (approved):
Thanks for pushing this

src/components/DpFlyout/DpFlyout.vue
```diff
@@ -2,13 +2,9 @@
   <span
     ref="target"
     v-on-click-outside="close"
-    class="o-flyout"
+    class="dp-flyout relative"
```
hwiem, line 5: Does this class (dp-flyout) exist?

src/components/DpResettableInput/DpResettableInput.vue
```diff
@@ -13,23 +13,25 @@
...
-        :size="iconSize" />
-    </button>
-    <!-- Slot for additional buttons -->
-    <slot />
+    <span class="space-x-0.5">
+      <button
+        v-if="!inputAttributes.disabled"
+        class="btn--blank o-link--default"
+        data-cy="resetButton"
+        :class="buttonClass"
```
hwiem, line 21: nitpick: `:class` should be above `data-cy`

hwiem (commented):
Very nice, thank you so much! :heart: Just a minor nitpick and a question

src/components/DpFlyout/DpFlyout.vue
spiess-demos: > Does this class (dp-flyout) exist?

this class (as well as the `dp-flyout-trigger` and so on) is intended to serve as a styling hook for "brute force overrides" for projects, where overriding css vars is not sufficient. It could be used in this way:

```
header > nav .dp-flyout {
  // Project specific stuff
}
```

I'd suggest to put it into all of our components, as a standard. We already discussed this with @salisdemos and while he suggested to use `data-cy` alternatively, i feel that using a real selector for this purpose is more appropriate.

hwiem: Thanks for explaining! I wonder if there's a good location to document this? Otherwise it might be confusing to find this class that is not defined anywhere

hwiem (approved):

src/components/DpFlyout/DpFlyout.vue
spiess-demos: totally agree - however i have no clue besides putting a paragraph in the welcome sections regarding component overrides.

hwiem: I think that's a good idea :+1:

src/components/DpResettableInput/DpResettableInput.vue
spiess-demos: 1bbba2a50ddc4d81e1f317858f4c8eed41baefaa

src/components/DpFlyout/DpFlyout.vue
spiess-demos: here we go be3889189ca1852d53eff83fa7ea4c80f6034463

PR #1271, opened 2025-05-09
fix(DPLAN-15601): DpTreeList. issue with sticky header

sakutademos:
**Description:** This PR fixes an issue with an error in stickier.js, when the Header or Footer measurements such as offsetHeight returns zero values. So create stickier, only if the element and its height exist.

- nextTick ensures that the code runs only after Vue has applied all reactive changes to the DOM

PR in Core https://github.com/demos-europe/demosplan-core/pull/4640


riechedemos (approved):
Thx lgtm!

PR #1275, opened 2025-05-13
Migrate to Tailwind CSS v4

spiess-demos:
## Summary
- Implement Tailwind CSS v4 compatibility while maintaining backward compatibility with v3
- Add CSS theme generation with `@theme` directive for Tailwind v4's CSS-first configuration
- Upgrade tailwindcss from v3.2.1 to v4.1.6 and related dependencies

## Key Changes
- **Token Generation**: Add new CSS output format for design tokens using Tailwind v4's `@theme` directive
- **CSS Variables**: Generate CSS variables in both `:root` and `@theme` blocks for maximum compatibility and namespacing
- **PostCSS Config**: Switch from JS to ESM format with updated plugins for Tailwind v4
- **Utility classes**: Add CSS utility classes for common needs (break-words, scrollbar-none, etc.)

While this implementation tries to maintain backward compatibility with v3, a few breaking changes still come with Tailwind v4 itself. For a detailed list, see https://tailwindcss.com/docs/upgrade-guide. These issues have already been tackled in this PR, regarding to demosplan-ui components. However, consuming applications should check their markup, too.


salisdemos (approved):
For me it looks good. seems like it works. 
Thanks for the effort!

PR #1278, opened 2025-05-16
fix: do not use var for breakpoints

spiess-demos:
Css vars do not work as values inside media queries.


salisdemos (approved):
Thanks. Legit

PR #1287, opened 2025-06-06
fix(refs DPLAN-12138): select nodes without mutating store state

hwiem:
**Ticket:** [DPLAN-12138](https://demoseurope.youtrack.cloud/issue/DPLAN-12138/Dokumentnamen-zeigen-NaN-undefined)

- selection of nodes is now extracted to a separate file (SelectionManager.js)
- instead of mutating store items directly, a separate object is used
- TreeList api does not change

To do:
- Test for DpTreeListNode still needs to be updated


salisdemos (approved):
still legit.Thank you

PR #1294, opened 2025-06-18
fix(refs DPLAN-15779): bump tiptap to version 3.0.0-beta.8

meissnerdemos:
### Ticket
[DPLAN-15779](https://demoseurope.youtrack.cloud/issue/DPLAN-15779)

This version bump is necessary for the BubbleMenu to function correctly. This version of TipTap no longer uses tippy.js as a dependency. It uses floating-ui instead, which also had to be bumped to a newer version. This is why the attributes of BubbleMenu are changed. 
Also, the new TipTap version now uses ES modules exclusively, so the webpack.config had to be changed.

### To Do
Investigate other impacts of this dependency change.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -60,7 +60,7 @@ const config = {
     rules: [
       {
         test: /\.(js|jsx)$/i,
-        include: ['@tiptap'].map(nodeModule => resolve('node_modules/' + nodeModule)),
+        include: ['@tiptap'].map(nodeModule => resolve('node_modules/' + nodeModule)), //@tiptap contains ES6+ code that needs transpilation for older browser support
         exclude: /node_modules/,
         loader: 'babel-loader'
       },


meissnerdemos (commented):
The new changes address the warning by adjusting the table import as well as an error caused by prosemirror. Tiptap imports different versions of prosemirror packages as dependencies which causes conflicts. This is addressed by adding the 'resolutions' section in the package.json that forces one consistent version.

webpack.config.js
```diff
@@ -60,6 +60,7 @@ const config = {
     rules: [
       {
         test: /\.(js|jsx)$/i,
+        include: ['@tiptap'].map(nodeModule => resolve('node_modules/' + nodeModule)),
```
salisdemos, line 63: is this still necessary?
It should not be. if there is a reason, we should add a comment to explain it.

salisdemos (changes requested):
:innocent:

webpack.config.js
meissnerdemos: addressed in e0666d623ee2988bf1103475e1b87fdbdc56a901

PR #1298, opened 2025-06-27
fix(DPLAN-2800): add new prop to the DpEditableList

sakutademos:
**Description:** This PR adds a new prop `closeOnSuccess` to the `DpEditableList` component:

- when using a dynamic `<component>` in our addons, it is not possible to call its methods via `refs`, so added an extra prop to enable closing the form.

PR in Addon: https://github.com/demos-europe/demosplan-addon-maillane/pull/183


salisdemos (approved):
legit. Thanks

PR #1310, opened 2025-07-10
chore: Migrate storybook to v9

salisdemos:
with storybook 9, they dropped the webpack support for vue. 
That's why we now have to use vite for storybook.

Some of the styles are broken, due to the circumstance that the components get styled from the consuming app. All components with an somehow broken style have to be migrated to tailwind in the End.

Can be tested with:
```
$ yarn
$ yarn build:tokens
$ yarn build:storybook
$ yarn sb
```

### DIFF
--- a/vite.config.mjs
+++ b/vite.config.mjs
@@ -1,0 +1,44 @@
+import { defineConfig } from 'vite'
+import vue from '@vitejs/plugin-vue'
+import path from 'path'
+
+export default defineConfig({
+  plugins: [vue()],
+  resolve: {
+    alias: {
+      '~': path.resolve(__dirname, 'src')
+    },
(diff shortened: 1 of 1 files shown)


meissnerdemos (commented):
image

I am getting this error when I try to build the storybook

src/components/DpButton/DpButton.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import type {Meta, StoryObj} from '@storybook/vue3'
+import type {Meta, StoryObj} from '@storybook/vue3-vite'
 import DpButton from './'
 import { SIZES as ICON_SIZES } from '~/components/DpIcon/util/iconConfig'
```
meissnerdemos, line 3: This line is throwing an error during the build

salisdemos:
strange. I dont get any errors. Can we have a look together :pray: ?

PR #1311, opened 2025-07-10
feat(refs DPLAN-16080): add optional search button to DpAutocomplete

hwiem:
**Ticket:** [DPLAN-16089](https://demoseurope.youtrack.cloud/issue/DPLAN-16080/Suchschlitz-nicht-schon-in-Mobilansicht)

- A new prop `searchButton` is added to DpAutocomplete
- if the prop is set to true, DpAutocomplete displays a search button attached to the autocomplete input on small screens, while also moving the placeholder outside of the input and displaying it above as a normal label. Inside the input, the placeholder 'search' is displayed instead
- this ensures that the placeholder in the input is not cut off on small screens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,16 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Added
+
 - ([#1311](https://github.com/demos-europe/demosplan-ui/pull/1311)) DpAutocomplete: Add optional search button ([@hwiem](https://github.com/hwiem))
+
+
+## v0.4.21 - 2025-07-15
+
+### Changed
+
+- ([#1303](https://github.com/demos-europe/demosplan-ui/pull/1303)) ActionMenu.js: improve keyboard navigation ([@meissnerdemos](https://github.com/meissnerdemos)
+
 
 ## v0.4.20 - 2025-07-01
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@demos-europe/demosplan-ui",
-  "version": "0.4.20",
+  "version": "0.4.21",
   "license": "MIT",
   "description": "Vue components, Vue directives, Design Token and Scss files to build interfaces for demosPlan.",
   "main": "./dist/demosplan-ui.umd.js",


meissnerdemos (approved):
Looks good to me and works locally, thank you!

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -1,37 +1,67 @@
...
         }">
         {{ placeholder }}
       </label>
+      <!-- Placeholder for small screens if search button is used -->
+      <template v-if="searchButton">
+        <span
+          :class="{ 'hidden': !isPlaceholderVisible }"
+          class="md:hidden h-full w-[80%] px-1 py-[2px] text-gray-400 flex items-center m-0">
+        {{ de.search.searching }}
+      </span>
```
sakutademos, line 45:
```suggestion
        </span>
```

src/components/DpAutocomplete/DpAutocomplete.vue
```diff
@@ -41,6 +71,15 @@
         @mousedown.stop.prevent="focusInput"
         @keydown.stop="runSpecialKeys"
         @focusout="isInputFocused = false" />
+      <dp-button
+        v-if="searchButton && !isPlaceholderVisible"
```
sakutademos, line 75: i am a bit confused about which button appears when, we have two `dp-button` components with the search icon and only their conditions differ. Maybe it could be refactored like this to remove the duplication:
```
<dp-button
  v-if="(isPlaceholderVisible || (searchButton && !isPlaceholderVisible))"
  ....other attributes...
  :class="isPlaceholderVisible ? 'border-2 border-l' : 'border-2 !border-l-1'"
 >
```

hwiem: Thanks, you're absolutely right e36907913e0bfea1d19584baab4f398ef40e2f9a

src/components/DpAutocomplete/DpAutocomplete.vue
hwiem: Thanks, done in 1c87298fe3fa2eb2e9e9325eda7b4188ff8a7743

PR #1312, opened 2025-07-11
chore: Change TS config moduleResolution to bundler

salisdemos:
otherwise we get some issues resolving esm only dependencies. and in the End this is the more correct setting anyway

for further details see
https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution-for-bundlers-typescript-runtimes-and-nodejs-loaders

And it fixes an issue where useSortable is not importable after bumping vueUse to 13.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,7 @@
     "jsx": "preserve",
     "allowJs": true,
     "checkJs": false,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "skipLibCheck": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "jsx": "preserve",
     "allowJs": true,
     "checkJs": false,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
hwiem, line 9: I don't know if I understand correctly the explanation in the link you added in the description, but doesn't it say that for libraries, you should not use `'bundler'`?

hwiem (approved):
Let's try it :)

PR #1313, opened 2025-07-11
feat(refs DPLAN-16085): Extend DpProgressBar to allow alternative app…

rafelddemos:
…earance


rafelddemos:
redone as 'feat' (instead of 'bug') in #1314

PR #1318, opened 2025-07-15
fix(DPLAN-16037): restore Vue 3 v-model for DpInput

sakutademos:
Ticket: [DPLAN-16037](https://demoseurope.youtrack.cloud/issue/DPLAN-16037) Nicht möglich um Berechtigte Absenderadressen für den E-Mail Import zu editieren oder löschen

**Description:**  In Vue 3 `v-model` on custom components no longer uses the native `value/input` API, it binds to a `modelValue` prop and emits `update:modelValue`. This change explicitly declares that mapping (using `defineOptions` in `<script setup>`) so that v-model works correctly on our custom inputs under compat mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


salisdemos:
are You sure `main` is Your target? Or should it be `0.4.x` ?
Otherwise legit and Thank You!

PR #1321, opened 2025-07-21
chore: Remove DpFormRow

salisdemos:
its useless. The same Result can be archived with just some tailwind classes


sakutademos (approved):
thx, lgtm!

PR #1323, opened 2025-07-22
fix(refs dplan-16160): Remove default submit behaviour and adjust styles

rafelddemos:
[DPLAN-16160](https://demoseurope.youtrack.cloud/issue/DPLAN-16160/Mitzeichnende-Toggle-ist-verrutscht-und-wenn-man-darauf-klickt-wird-das-STN-Formular-submitted)

This PR adjusts the changes made in the following PR:
https://github.com/demos-europe/demosplan-ui/pull/1269/files

- it re-adds the type of 'button' to remove the default submit behaviour when DpAccordion is inside a form
- adjusts Tailwind classes

Related PR in demosplan-core: https://github.com/demos-europe/demosplan-core/pull/4963
(Adds padding left in the side menu --> with the new position of the caret icon (after the text), the text in the side menu was too far left)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Changed
+- ([#1318](https://github.com/demos-europe/demosplan-ui/pull/1318)) DpEditableList: ensure Vue 3 v-model works correctly under compat mode ([@sakutademos](https://github.com/sakutademos)
 - ([#1321](https://github.com/demos-europe/demosplan-ui/pull/1321)) BREAKING: Remove DpFormRow ([@salisdemos](https://github.com/salisdemos))
 - ([#1323](https://github.com/demos-europe/demosplan-ui/pull/1323)) DpAccordion: Remove default submit behaviour and adjust styles ([@rafelddemos](https://github.com/rafelddemos))
-
-### Changed
-- ([#1318](https://github.com/demos-europe/demosplan-ui/pull/1318)) DpEditableList: ensure Vue 3 v-model works correctly under compat mode ([@sakutademos](https://github.com/sakutademos)
 
 
 ## v0.5.3 - 2025-07-17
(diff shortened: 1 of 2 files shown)


src/components/DpAccordion/DpAccordion.vue
```diff
@@ -3,8 +3,9 @@
     <button
       v-if="title !== ''"
       :aria-expanded="isVisible.toString()"
-      class="flex items-center justify-between w-full hover:text-interactive hover:cursor-pointer active:text-interactive-hover"
       :data-cy="dataCy"
+      class="flex items-center gap-2 text-interactive hover:text-interactive-hover hover:cursor-pointer"
```
hwiem, line 7: I am not 100% sure how the icon should be positioned, but according to the [accordion designs](https://www.figma.com/design/Kj7EYQ8gCDFZwoVbbakEx7/Components--v1-?node-id=1242-667&p=f&t=oMEmZyoW3nC2KK6y-0) the text should have the default color by default (meaning we shouldn't use `text-interactive`)

rafelddemos: or maybe I could try to override the color setting only for the side menu..

rafelddemos: adjusted the colors according to the figma design in 20ceab7a25800b90bad678bd6f50b0ccce94aa28

Not sure about the positioning of the caret inside forms and maybe-bottom-border - I would leave full-with for now and open a new ticket for this issue when we have the final UI-feedback on that. What do you think?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ### Changed
 - ([#1321](https://github.com/demos-europe/demosplan-ui/pull/1321)) BREAKING: Remove DpFormRow ([@salisdemos](https://github.com/salisdemos))
+- ([#1323](https://github.com/demos-europe/demosplan-ui/pull/1323)) DpAccordion: Remove default submit behaviour and adjust styles ([@rafelddemos](https://github.com/rafelddemos))
```
hwiem, line 9: I think this maybe happened during a merge, but there is already a ### Changed block under ## Unreleased, and we should move these two entries to that block (below)

hwiem (approved):
Nice! ❤️

PR #1324, opened 2025-07-24
B dplan 16159 fix search field

meissnerdemos:
### Ticket
[DPLAN-16159](https://demoseurope.youtrack.cloud/issue/DPLAN-16159)

This PR adds a prop to control whether the DpFlyout trigger has position: relative or absolute and fixes the logic for detection whether DpResettableInput has meaningful slot content. Returning 'false' for fragments was incorrect since these can contain meaningful nodes in Vue.

Related PR in Core: [4968](https://github.com/demos-europe/demosplan-core/pull/4968)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,14 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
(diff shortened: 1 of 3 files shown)


hwiem (approved):
Lgtm, thank you!

PR #1325, opened 2025-07-25
fix(refs DPLAN-16172): use tailwind classes to apply correct styling

hwiem:
The stylings for form components as defined in our design system are reflected in our tailwind classes, so we should use them (before, styling was defined in demosplan custom classes). if we want to override these styles in a project, we need to override the css var(s) associated with the tailwind class (--dp-border-color-input-disabled in this case) in the project (color) settings.


rafelddemos (approved):
lgtm, thank you :) code-review only.

PR #1328, opened 2025-07-31
fix(refs DPLAN-16159): add more prosemirror packages to resolutions

meissnerdemos:
### Ticket
[DPLAN-16207](https://demoseurope.youtrack.cloud/issue/DPLAN-16207)

This PR adds more prosemirror packages to the resolutions in the package.json. The error in the ticket was caused by different versions of prosemirror packages being imported, since some of them have other prosemirror packages as dependencies themselves, and it appears that not all versions are compatible with each other. The "resolutions" section of the package.json is referenced by yarn to resolve these version conflicts and ensure that only one version is imported. 

### How to test
1. Login as admin in a project that has segments and navigate to the 'Stellungnahmen aufteilen' section.
2. Start to create a segment, add a tag to it, and then abort the process with the abort button.
3. Test if there is any weird behavior now like being unable to make another segment (if there is, the bug is still there).
4. Also check the console for the error `Uncaught TypeError: style.eq is not a function`

### Related PR
#1294 Here a different error also involved conflicting prosemirror package versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
+- ([#1328](https://github.com/demos-europe/demosplan-ui/pull/1328)) Add more prosemirror packages to resolutions to prevent dependency conflicts ([@meissnerdemos](https://github.com/meissnerdemos))
+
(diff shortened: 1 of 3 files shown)


rafelddemos (approved):
lgtm and works, no console errors. thanks! :)

the Ticket number in the title and the PR description should probably be 16207 though?

PR #1330, opened 2025-08-04
feat(refs DPLAN-14969, #AB23560): Allow non-error DpNotification MSGs to persist

rafelddemos:
[DPLAN-14969](https://demoseurope.youtrack.cloud/issue/DPLAN-14969/ADO-Issue-23560-Rechtzeitige-Ankundigung-Ausloggen-damit-gespeichert-werden-kann)

So far, only the notifications of type 'error' were not set to disappear after 20 seconds.
This PR expands the condition for the DpNotification setTimeout, so that other types of notifications (warning, info) can also be allowed to persist, if needed.
'message.persist' is implemented as optional (and defaults to 'false'), so that already existing implementations are not affected.

Related core-PRs: 
FE: https://github.com/demos-europe/demosplan-core/pull/4970
BE: https://github.com/demos-europe/demosplan-core/pull/4959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Changed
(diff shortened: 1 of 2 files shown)


hwiem (approved):
Nice :)

PR #1332, opened 2025-08-08
fix(refs DPLAN-16247): remove relative positioning on focus

hwiem:
Ticket: [DPLAN-16247](https://demoseurope.youtrack.cloud/issue/DPLAN-16247/Originalstellungnahmen-Nach-dem-Klick-auf-Exportieren-springt-der-Button-nach-links)

Because of the `focus:relative`, DpButton occasionally moved to a different position in the interface on focus. I'm sure it has something to do with the tailwind update. But since simply removing the `focus:relative` style does not break anything (as far as I could tell), it seemed to be the easiest solution to do just that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Fixed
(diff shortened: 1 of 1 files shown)


rafelddemos (approved):
lgtm and seems to do the trick, thanks :)

I guess, it solves this [Bug](https://demoseurope.youtrack.cloud/issue/DPLAN-16119/Seitenanfang-Button-springt-weg) as well, afaiks :)

PR #1340, opened 2025-08-25
feat: add tests for DpCheckboxGroup

meissnerdemos:
This PR adds unit tests for DpCheckboxGroup.

I also discovered a bug in DpCheckboxGroup: On mount, selected was set to false for every option, regardless of whether 'true' was passed via the selectedOptions prop. I included a fix in this PR.

### DIFF
--- a/tests/form/DpCheckboxGroup.spec.js
+++ b/tests/form/DpCheckboxGroup.spec.js
@@ -1,5 +1,5 @@
 import { shallowMount } from '@vue/test-utils'
-import DpCheckboxGroup from '~/components/DpCheckboxGroup/DpCheckboxGroup.vue'
+import DpCheckboxGroup from '~/components/DpCheckboxGroup/DpCheckboxGroup'
 
 describe('DpCheckboxGroup', () => {
   let wrapper
@@ -21,7 +21,7 @@ describe('DpCheckboxGroup', () => {
     })
   })
 
-  it('initializes selected state with all options set to false', () => {
+  it('initializes selected state with all options set to false if selectedOptions is empty', () => {
     expect(wrapper.vm.selected).toEqual({
       option1: false,
       option2: false,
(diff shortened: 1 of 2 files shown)


tests/form/DpCheckboxGroup.spec.js
```diff
@@ -0,0 +1,118 @@
+import { shallowMount } from '@vue/test-utils'
+import DpCheckboxGroup from '~/components/DpCheckboxGroup/DpCheckboxGroup.vue'
```
hwiem, line 2:
```suggestion
import DpCheckboxGroup from '~/components/DpCheckboxGroup/DpCheckboxGroup'
```

tests/form/DpCheckboxGroup.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+    expect(wrapper.emitted('update')).toBeTruthy()
+    expect(wrapper.emitted('update')[0]).toEqual([{
+      option1: true,
+      option2: false,
+      option3: false
+    }])
+  })
+
+  it('emits update event with correct merged state', async () => {
```
hwiem, line 56: Does this test case really test anything that wasn't tested in the test case before it?

tests/form/DpCheckboxGroup.spec.js
```diff
@@ -0,0 +1,118 @@
...
+      props: {
+        options: defaultOptions,
+        label: 'Test Legend',
+        inline: false,
+        dataCy: 'test-group'
+      }
+    })
+  })
+
+  it('initializes selected state with all options set to false', () => {
```
hwiem, line 24:
```suggestion
  it('initializes selected state with all options set to false if selectedOptions is empty', () => {
```

hwiem (approved):
Thank you for adding this :heart: Just nitpicks

tests/form/DpCheckboxGroup.spec.js
meissnerdemos: You're right, it doesn't actually matter which checkbox is selected. Removed in 01b0b165c67d61879179f83997272d00b57c5207

PR #1362, opened 2025-10-08
chore: enable new eslint rules

hwiem:
- since we agreed to enable these rules in demosplan, we should also do so in demosplan-ui
- trailing commas should now always be added on multi-line objects, arrays, imports, etc.
- closing brackets should be placed on a new line for multi-line elements


huellnerdemos (approved):
Wow, a lot of trailing commas and line breaks :D Looks nice, thank you!

PR #1456, opened 2026-03-18
chore: Add CSRF token to Api 2.0 defaut headers

rafelddemos:
This PR adds X-CSRF token to the default headers of api2defaultHeaders.

Without it, `CsrfSubscriber` fires a dev notification (error.csrf.missing) on the backend for every non-GET api2 request, because the header is entirely absent.                                                                                                    
                                                                                                                       
Note: Handling this centrally in DpApi.js means no per-component CSRF header management is needed — any api2 call made through dpApi will include the token automatically.


mussbach (approved):
lgtm, thx

PR #1470, opened 2026-04-16
fix: add font weight classes from tailwind

meissnerdemos:
The project imports a custom theme (`styles/theme.css`) instead of `tailwindcss/theme.css`. The custom theme defines colors, spacing, font sizes etc. but was missing font weight variables.

In Tailwind 4, utilities like `font-semibold` or `font-bold` are only generated when the corresponding `--font-weight-*` CSS variable exists in the theme. Without these variables, `font-semibold` produces a class in the HTML but no matching CSS rule, so the declaration silently has no effect.

This adds all nine default font weight values from the Tailwind 4 default theme.

## How to review/test

- Verify `font-semibold` works (for example in StatementPublicationAndVoting in the 'Details' section of a STN or in the mentioned PR)

## Related PR
fix is needed for this PR in core: [#6091](https://github.com/demos-europe/demosplan-core/pull/6091)

### DIFF
--- a/styles/theme.css
+++ b/styles/theme.css
@@ -1,6 +1,16 @@
 @import "../tokens/dist/css/theme.css";
 
 @theme {
+  --font-weight-thin: 100;
+  --font-weight-extralight: 200;
+  --font-weight-light: 300;
+  --font-weight-normal: 400;
+  --font-weight-medium: 500;
+  --font-weight-semibold: 600;
+  --font-weight-bold: 700;
+  --font-weight-extrabold: 800;
+  --font-weight-black: 900;
+
   --flex-shrink-2: 2;
   --animate-busy: busy 1.5s linear infinite;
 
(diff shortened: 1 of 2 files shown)


styles/theme.css
```diff
@@ -1,6 +1,16 @@
 @import "../tokens/dist/css/theme.css";
 
 @theme {
+  --font-weight-thin: 100;
```
hwiem: I think we don't need all of these - as far as I can tell, we only use font-weights 400, 500, 600 and 700 at the moment, so I would add only those. If at some point we need another value, we can still add it then.

hwiem (commented):
Thanks so much for fixing this! ❤️ Just one concern

styles/theme.css
meissnerdemos: Thanks for looking it up! I removed the extra weights in 8e36e9097dbbee57839e747581b11be7a1f6e853

hwiem (approved):
Lgtm, thank you! ❤️

PR #1471, opened 2026-04-17
fix: resolve form-data to patched version

meissnerdemos:
The transitive dependency `form-data@4.0.0` (pulled in via `jest-environment-jsdom` → `jsdom`) uses an unsafe random function for multipart boundaries ([Dependabot alert 86](https://github.com/demos-europe/demosplan-ui/security/dependabot/86)). Since it's not a direct dependency, a `resolutions` entry forces Yarn to resolve `^4.0.0` to `4.0.4`.

### How to review/test
1. Run `yarn install` - verify no resolution conflicts
2. Run `yarn test` - test suite confirms nothing breaks

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,7 @@
     "qs": "^6.14.2",
     "prosemirror-view": "1.40.1",
     "prosemirror-state": "1.4.3",
-    "prosemirror-transform": "1.10.4"
+    "prosemirror-transform": "1.10.4",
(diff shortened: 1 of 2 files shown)


hwiem (approved):
Lgtm, thank you! ❤️ 
This reminds me that I created a skill to check if any of the resolutions are outdated (because we need to make sure that we don't keep a version pinned there forever), and I should publish it. :)

PR #1476, opened 2026-04-27
Add unit test for DpIcon

riechedemos:
This PR adds a unit test for DpIcon.


tests/DpIcon.spec.js
```diff
@@ -0,0 +1,31 @@
+import DpIcon from '~/components/DpIcon/DpIcon.vue'
```
hwiem: nitpick
```suggestion
import DpIcon from '~/components/DpIcon/DpIcon'
```

meissnerdemos (changes requested):
Thank you for working on this! Unfortunately, currently these tests only test that vue correctly passed the props, and nothing specific to DpIcon.
I would recommend that you test the attributes, not the props. For example:

```
it.each([                                                                     
    ['small', '16'],
    ['medium', '20'],
    ['large', '24'],
    ['xlarge', '32'],
  ])('maps size "%s" to %s pixels', (size, expected) => {
    const wrapper = shallowMount(DpIcon, { props: { icon: 'check', size } })    
    expect(wrapper.attributes('size')).toBe(expected)
  }) 
```

You can look at `iconConfig.ts` to find the values that are passed to the template. I would test the sizes, weigths and proportions. And maybe also test the default values, so what happens when you give no props.

riechedemos:
Thank you that makes way more sense :D 106f850b919ed79a15ec4e075d86f596ad0571c9

tests/DpIcon.spec.js
riechedemos: changed it a7bbbc3396fd291906aed6c7475ceadc66b63d08

tests/DpIcon.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        props: { icon: 'check', weight },
+      })
+      expect(wrapper.attributes('weight')).toBe(weight)
+    })
+
+  it.each([
+    ['arrow-up', 'portrait'],
+    ['check', 'landscape'],
+    ['plus', 'square'],
+  ])('applies proportion class "%s" for icon "%s"', (icon, proportionClass) => {
```
hwiem: Shouldn't it be
```
['portrait', 'arrow-up' ],
[ 'landscape', 'check'],
['square', 'plus'],
```
?

hwiem (commented):
Just one small issue, rest looks good :)

tests/DpIcon.spec.js
riechedemos: ups 7c11ab456b4ee2928da8022e5f15f9931778687c

meissnerdemos (approved):
Lgtm and works! Thank you for working on this 🙂

PR #1477, opened 2026-04-27
refactor(refs DPLAN-16770): migrate from jest to vitest

hwiem:
**Ticket:** [DPLAN-16770](https://demoseurope.youtrack.cloud/issue/DPLAN-16770/Migration-von-jest-zu-vitest)

### Why migrate to vitest?
The @vue/vue3-jest package is no longer maintained, leading to compatibility issues with newer versions of jest dependencies, so we can no longer update them. Also, vue officially recommends using vitest as a testing framework for vue3.

### Additional fixes in this PR

- add a `null` check to prevent calling `getComputedStyle` on `null`
- define time zone (env.TZ) in vite config so tests pass in UTC CI environment

### How to review/test

- from `local_modules/demosplan-ui` in the demosplan container, run `yarn test` -> all tests should run successfully
- run `yarn test:watch` to test vitest ui (browser ui that shows the results of all test cases); in order for this to work from inside the container, you need to add these ports to your local docker-compose.yml in demosplan:
```
ports:
      - '51204:51204'                                                                              
      - '51205:51205'
```
- run `yarn test:coverage` to confirm it works

 ### PR Checklist                                                                                     
   
  - [x] Run `yarn test`                                                                                  
  - [x] Run `yarn lint`
  - [x] Update changelog


src/components/DpTooltip/utils/tooltip.js
```diff
@@ -24,22 +24,30 @@ const destroyTooltip = (wrapperEl) => {
...
 const getZIndex = (element) => {
-  const z = window.getComputedStyle(element).getPropertyValue('z-index')
-  if (isNaN(z)) {
+  if (!element) {
+    return 1
+  }
+
+  const z = globalThis.getComputedStyle(element).getPropertyValue('z-index')
+
+  if (Number.isNaN(z)) {
```
elmasdemos: I might be misunderstanding this, but z is always a string (since getComputedStyle(...).getPropertyValue('z-index') returns a string). Because of that, Number.isNaN(z) would always evaluate to false, so this logic would never run. Am I right?

tests/__snapshots__/DpNotification.spec.js.snap
```diff
@@ -1,24 +1,34 @@
...
+"<div class="c-notify__message c-notify__message--error" role="alert"><button class="c-notify__closer" type="button" aria-label="Hinweis ausblenden"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 256" width="20" height="20" fill="currentColor" class="square c-notify__icon" aria-hidden="true">
+      <g>
+        <path d="M128,24A104,104,0,1,0,232,128,104.11,104.11,0,0,0,128,24Zm37.66,130.34a8,8,0,0,1-11.32,11.32L128,139.31l-26.34,26.35a8,8,0,0,1-11.32-11.32L116.69,128,90.34,101.66a8,8,0,0,1,11.32-11.32L128,116.69l26.34-26.35a8,8,0,0,1,11.32,11.32L139.31,128Z"></path>
+      </g>
+    </svg></button>
+  <div class="flow-root"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 256" width="20" height="20" fill="currentColor" class="square c-notify__icon mt-px mr-1 float-left" aria-hidden="true">
+      <g>
+        <path d="M128,24A104,104,0,1,0,232,128,104.11,104.11,0,0,0,128,24Zm-8,56a8,8,0,0,1,16,0v56a8,8,0,0,1-16,0Zm8,104a12,12,0,1,1,12-12A12,12,0,0,1,128,184Z"></path>
+      </g>
+    </svg>
```
elmasdemos, line 13: Could you explain why you made this change? It looks like there was no rendered content before and now the SVG is included in the snapshot.

tests/utils/debounce.test.js
```diff
@@ -5,45 +5,45 @@ describe('debounce', () => {
...
     expect(callback).toBeCalledTimes(1)
 
-    jest.advanceTimersByTime(1)
+    vi.advanceTimersByTime(1)
     expect(callback).toBeCalledTimes(1)
 
-    jest.advanceTimersByTime(700)
+    vi.advanceTimersByTime(700)
     debounced()
     expect(callback).toBeCalledTimes(2)
```
elmasdemos, line 44: I refactored the debounce tests to focus on behavior rather than timer mechanics. This should make them easier to read and more understandable.
```
it('does not call the function if the debounce is reset within the cooldown time', () => {
 const callback = vi.fn()
  const debounced = debounce(callback, 500)

  debounced()
  vi.advanceTimersByTime(400)

  debounced() // reset

  vi.advanceTimersByTime(500)
  expect(callback).toBeCalledTimes(1)
})

it('fires immediately and ignores calls within the cooldown time', () => {
  const callback = vi.fn()
  const debounced = debounce(callback, 500, true)

  debounced()
  expect(callback).toBeCalledTimes(1)     //immediate execution

  debounced()
  vi.advanceTimersByTime(100)          //still in cooldown (400ms left)
  expect(callback).toBeCalledTimes(1)

  vi.advanceTimersByTime(500)        //cooldown finished (total 600ms)
  debounced()
  expect(callback).toBeCalledTimes(2)
})
```

src/components/DpTooltip/utils/tooltip.js
hwiem: Thanks for spotting this, I just applied a sonarcloud suggestion. Fixed in 9e690dcee1796deca911a3816034f1d573019c64

tests/__snapshots__/DpNotification.spec.js.snap
hwiem: After migrating to vitest, when I ran the tests some snapshots needed to be updated. Since we're using `mount` in the DpNotification test, the icon from DpIcon should indeed be included in the snapshot, so for me the real question is why wasn't that the case before with the jest setup (I don't have an answer for that).

tests/utils/debounce.test.js
hwiem: I consider it outside of the scope of this PR, because the goal of the PR is to migrate to vitest, not to refactor any tests more than necessary. But since you already put in the effort, and I agree it's an improvement, I integrated your suggestion: 2da829e8f8dda16a8684316161f30e50ad1a5298 (and I appreciate your thorough review ❤️ )

elmasdemos (approved):
thanks, lgtm <3

PR #1478, opened 2026-04-27
fix(refs DPLAN-17692): restore beforeList slot for single selects

hwiem:
- if `multiple` was `false`, the `beforeList` slot in DpMultiselect was no longer displayed; but the slot is not only used for `selectionControls` (which only make sense if `multiple` is `true`), but also to display other content
- now the slot is displayed either if the parent component provides content for it or if `selectionControls` and `multiple` are `true`


rafelddemos (approved):
lgtm (code review), thank you :)

PR #1481, opened 2026-04-28
fix: add floor to tooltip z index

meissnerdemos:
### Description
The tooltip util sets an inline `z-index = getZIndex(trigger) + 1`, walking the trigger's ancestor chain for the nearest numeric z-index. When no ancestor has one (e.g. the tooltip lives inside a native `<dialog>`, which relies on the top layer rather than z-index), the fallback resolves to `1` and the tooltip ends up at z-index `2`. Sibling elements with even small explicit z-index utilities (such as `DpSearchField`'s search button at `z-[5]`) then overlap the tooltip.

### Fix
Floor the inline z-index at `2000`, matching the `tooltip` design token (`tokens/src/zIndex.json`). The heuristic still lifts the tooltip above containers with higher numeric z-index (e.g. `DpFlyout` at `z-flyout = 2100`), but never falls below the documented tooltip layer.

### How to test
Trigger a tooltip on an element inside a `DpModal` that has a sibling with a small explicit z-index (e.g. RecommendationModal). The tooltip should appear above the sibling.


hwiem (changes requested):
Code lgtm, but a changelog entry is missing

meissnerdemos:
> Code lgtm, but a changelog entry is missing

Sorry, I was rushing :see_no_evil: added the missing entry in 0a3a8d97ec64f1dd9b0663b83ea9ed467e7de048

hwiem (approved):
Lgtm!

PR #1483, opened 2026-04-29
chore: remove obsolete resolutions

hwiem:
- form-data was removed as a transitive dependency when migrating from jest to vitest
- qs was pinned because @vue/cli-service required a vulnerable version, but @vue/cli-service was removed as a dependency when migrating from webpack to vite


meissnerdemos (approved):
Thanks!

PR #1485, opened 2026-05-08
feat(ref. DPLAN-17737) Change session to localStorage DpDataTable

riechedemos:
### Ticket                                            
  [DPLAN-17737](https://demoseurope.youtrack.cloud/issue/DPLAN-17737)                       
                                                                                                    
Add `columnWidthStorageKey` prop on `DpDataTable` so consumers can opt in to persisting column widths in `localStorage` instead of `sessionStorage`. The prop is forwarded through `DpTableHeader` to `DpResizableColumn`. When the prop is unset, the existing `sessionStorage`-based behavior is preserved unchanged.                                           
                                                                                                    
  ### How to review/test                                                                            
-Go to SegmentsList (Abschnitte) → resize columns, close the tab, reopen → widths are restored                                                                    
                                                                                                    
  ### Linked PRs                                                                                    
[demos-europe/demosplan-core#<PR-Nummer>](https://github.com/demos-europe/demosplan-core/pull/6239)

### DIFF
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -755,8 +755,10 @@ export default {
     },
 
     readStoredColWidth (headerField) {
-      // When columnWidthStorageKey is set, persist column widths in localStorage (survives tab close).
-      // Otherwise, fall back to the legacy sessionStorage behavior used by other DpDataTable consumers.
+      /*
+       * When columnWidthStorageKey is set, persist column widths in localStorage (survives tab close).
+       * Otherwise, fall back to the legacy sessionStorage behavior used by other DpDataTable consumers.
+       */
       if (this.columnWidthStorageKey) {
         try {
           const stored = localStorage.getItem(`dpDataTable:colWidth:${this.columnWidthStorageKey}:${headerField}`)
--- a/src/components/DpDataTable/DpDataTable.vue
+++ b/src/components/DpDataTable/DpDataTable.vue
@@ -27,6 +27,7 @@
           :data-cy="`${dataCy}:header`"
           :all-expanded="allExpanded"
           :checked="allSelected"
+          :column-width-storage-key="columnWidthStorageKey"
           :density="density"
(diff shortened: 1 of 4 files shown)


src/components/DpDataTable/DpDataTable.vue
huellnerdemos, line 582: When i tested changing the column width, it was stored in the local storage  correctly. But after reloading the page, the columns still was there, but the width was set back to default values. The correct values where still saved in the storage, but something had overwritten it in the style width. I investigated a bit and found out, that a small change at two places can fix this. It's just changing the order of some method calls.  
```suggestion
           this.fillContainerWidth(tableHeaderElements)
           this.setColsWidth(tableHeaderElements)
```

huellnerdemos, line 969: At this place we maybe also have to change the order of some method calls, to prevent `this.fillContainerWidth(tableHeaderElements)` to override the values that came from the storage through `this.setColsWidth(tableHeaderElements)` before.


```suggestion

     this.tableEl.style.tableLayout = 'fixed'
     this.tableEl.classList.add('is-fixed')

     this.fillContainerWidth(tableHeaderElements)
     this.setColsWidth(tableHeaderElements)
     
```

huellnerdemos (changes requested):
Thank you! 🎊 🎉  Code looks very good. I tested it and a lot of it worked fine, but when i reloaded the page after adding some columns and change the width of them, they switched back to a default width. I checked the local storage and the values were still saved correctly. So it seemed that something overrides the values after they came from the storage. I added two comments with suggestions, that worked for me. Maybe that could be a solution.

src/components/DpDataTable/DpDataTable.vue
riechedemos: thank you 4c1bf79c2a30b64537857c277f3848eaed533bb3

huellnerdemos (approved):
Code looks good. Tested it and it worked well now. Thank you 🎉

PR #1488, opened 2026-05-13
feat(refs: DPLAN-17223) manual tag ordering within topics in TagsList

riechedemos:
## Ticket
[DPLAN-17223](https://demoseurope.youtrack.cloud/issue/DPLAN-17223/Ado-issue-48739-Sortierte-Schlagwortliste)

## Summary

Two fixes for cross-list drag-and-drop in `DpTreeList`, one of which is in
`DpDraggable` (which `DpTreeList` wraps internally).

### Dragged element correctly identified at `@end` — `DpDraggable.vue`

When dragging a tag from one topic to another, the `@end` handler received
an incorrect `item` — the value at the original `oldIndex` in the (now
re-indexed) source list, not the actually dragged element. `onEnd` now uses
`e.item` from SortableJS directly.

### `end` and `start` declared in emits — `DpTreeList.vue`, `DpTreeListNode.vue`

Both events were bubbled by `handleDrag(...)` but missing from the `emits`
arrays. Consumers binding to either saw Vue warnings. Added both to the arrays.

## Context

Driven by DPLAN-17223 — consumer needs reliable cross-list drop handling to
persist tag moves at the dropped position.

## How to test

1. Use `DpTreeList` with multiple topics -> TagsList.vue
2. Drag a leaf from one topic to another
3. In the `@end` handler, log the `item` (3rd arg) → should be the dragged
   element, not a sibling from the source branch
4. Console should be free of Vue warnings about undeclared `end` / `start` events

## Linked PRs
https://github.com/demos-europe/demosplan-core/pull/6233

PR #1489, opened 2026-05-18
fix(refs DPLAN-17223): clean up orphaned DOM node on cross-list drop

riechedemos:
## Ticket
[DPLAN-17223](https://demoseurope.youtrack.cloud/issue/DPLAN-17223/Ado-issue-48739-Sortierte-Schlagwortliste)

This PR adds a bugfix to https://github.com/demos-europe/demosplan-ui/pull/1488

SortableJS moves the dragged element directly in the DOM; Vue's v-for then rebuilds the destination from state and leaves the moved node behind as a duplicate. Remove it in onEnd so Vue is the single source of truth.


huellnerdemos (approved):
Tested it and it works fine. Thank you 🎉

PR #1496, opened 2026-05-26
fix: use new plyr import path

meissnerdemos:
Fix build error from new Plyr version: replace removed deep import plyr/dist/plyr.polyfilled.min with main entry plyr.


hwiem (approved):
Thanks!

PR #1501, opened 2026-06-02
feat(refs DPLAN-17527): Add prop for invalid state to DpInput

rafelddemos:
This PR adds a prop for invalid styling to `DpInput` that can be used for client-validation like duplicate values depending on the component's state. It pushes the class to the input's `classes` array.

It also adds aria-invalid attribute to `DpInput` and the prop check to the unit test.

Related Ticket:[ DPLAN-17527](https://demoseurope.youtrack.cloud/issue/DPLAN-17527/NEU-ADO-50748-Bearbeiten-von-custom-Verfahrensschritten)


src/components/DpInput/DpInput.vue
```diff
@@ -309,6 +321,10 @@ const classes = computed(() => {
     _classes.push('text-input bg-surface border border-input cursor-text')
   }
 
+  if (props.showErrorBorder) {
+    _classes.push('border-status-failed')
+  }
```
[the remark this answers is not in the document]
rafelddemos: added in 6a049f13423c164ba6e044a6e0750b5a2ef7618a

src/components/DpInput/DpInput.vue
```diff
@@ -221,6 +221,18 @@ const props = defineProps({
     validator: (prop: string) => ['full', 'left', 'right'].includes(prop),
   },
 
+  /**
+   * When true, `border-status-failed` class is added to `<input>` classes,
+   * rendering the same thin error border as a native required-empty field.
+   * This is independent of the dp-validate system,
+   * e.g. for client-side checks like duplicate values in the input.
+   */
+  showErrorBorder: {
```
hwiem: I'd suggest to rename the prop to `invalid` because that expresses more accurately what the prop represents and what it should be used for; if we add the focus style and the aria-invalid attribute, `showErrorBorder` is no longer accurate.

src/components/DpInput/DpInput.vue
```diff
@@ -309,6 +321,10 @@ const classes = computed(() => {
     _classes.push('text-input bg-surface border border-input cursor-text')
   }
 
+  if (props.showErrorBorder) {
+    _classes.push('border-status-failed')
```
hwiem: To make this accessible, we should also add an `aria-invalid` attribute to the input if `invalid`/`showErrorBorder` is `true`; that way, screen readers will announce the input as invalid.

To make the styling consistent with dp-validate, we should also add `focus-visible:outline-interactive-warning`

src/components/DpInput/DpInput.vue
```diff
@@ -221,6 +221,18 @@ const props = defineProps({
     validator: (prop: string) => ['full', 'left', 'right'].includes(prop),
   },
 
+  /**
+   * When true, `border-status-failed` class is added to `<input>` classes,
+   * rendering the same thin error border as a native required-empty field.
+   * This is independent of the dp-validate system,
+   * e.g. for client-side checks like duplicate values in the input.
+   */
```
hwiem, line 229: I think we don't need to explicitly state here which styles are applied, and the reference should not be native styling, but our design system. So I'd change the comment to something like
```
/**
 * When true, invalid styling and `aria-invalid` are applied to the input.      
 * Use for client-side, state-dependent checks that dp-validate
 * can't express (e.g. a value that must be unique across sibling rows).
*/
```

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 
 ## UNRELEASED
 
+### Added
+- ([#1501](https://github.com/demos-europe/demosplan-ui/pull/1501)) DpInput: Add prop for error border style ([@rafelddemos](https://github.com/rafelddemos))
```
hwiem: Maybe something like "Add prop for invalid state" is more accurate?

hwiem (changes requested):

CHANGELOG.md
rafelddemos: adjusted in 38a369473a416fd59b2e9902c47e2d7b398848a3

src/components/DpInput/DpInput.vue
rafelddemos: adjusted in 38a369473a416fd59b2e9902c47e2d7b398848a3

src/components/DpInput/DpInput.vue
rafelddemos: yes, good idea, thanks for the input! I made the changes in 38a369473a416fd59b2e9902c47e2d7b398848a3

src/components/DpInput/DpInput.vue
rafelddemos: I did not manage to achieve the shadowy look from the figma design for this PR, but moved the `focus-visible:outline-interactive` class to only be applied to valid elements to get rid of the overlapping color styles.

hwiem (approved):
Lgtm, thank you for updating the unit test ❤️

PR #1509, opened 2026-06-08
chore: disable install scripts in dependencies

hwiem:
- yarn no longer runs preinstall/postinstall scripts from dependencies, closing the install-time arbitrary-code-execution vector (npm worm protection)
- skipped build scripts (core-js, esbuild, tailwindcss/oxide) are safe to ignore (core-js: just a funding message, esbuild/oxide: platform binaries don't have to be built or fetched via script, they come in as separate dependencies)


mussbach (approved):
thank you!

PR #1514, opened 2026-06-10
Feat(DPLAN-17486): add a new helper function `reformatDateString`

sakutademos:
**Ticket:** https://demoseurope.youtrack.cloud/issue/DPLAN-17486/Datumsfeld-zur-Fristsetzung-bei-Bearbeiterzuweisung

**Description:** Add a new helper function `reformatDateString`, as we need it in several cases where dates must be converted from ISO format to a standard format and vice versa.

- add unit tests for the new `reformatDateString()` function
- refactor `toDate()` function and fix unit tests

**PR in Core:** https://github.com/demos-europe/demosplan-core/pull/6283


meissnerdemos (changes requested):
Overall, the code looks good! However, I think since we already have unit tests for `date.js`, it would be good to extend them for this new function.
Also, using `dayjs` with the German date format requires the `CustomParseFormat` plugin (see [here](https://day.js.org/docs/en/parse/string-format)). Its already being loaded in `DpDateTimePicker`, so for now it still ends up in the bundle and can be used by `date.js`. I think it would be safer to also add it in `date.js` in case we ever change `DpDateTimePicker`. 
That would mean adding this line somewhere appropriate:
`dayjs.extend(customParseFormat)`

sakutademos:
> Overall, the code looks good! However, I think since we already have unit tests for `date.js`, it would be good to extend them for this new function. Also, using `dayjs` with the German date format requires the `CustomParseFormat` plugin (see [here](https://day.js.org/docs/en/parse/string-format)). Its already being loaded in `DpDateTimePicker`, so for now it still ends up in the bundle and can be used by `date.js`. I think it would be safer to also add it in `date.js` in case we ever change `DpDateTimePicker`. That would mean adding this line somewhere appropriate: `dayjs.extend(customParseFormat)`

totally agree, we neeed tests!!! :)) so tests were added as well as CustomParseFormat plugin. The toDate() function was also refactored as part of this PR

meissnerdemos (approved):
Really nice ✨ Thank you for putting in so much effort into improving all these files ❤️

PR #1542, opened 2026-07-24
chore: bump brace-expansion to patched version

hwiem:
fixes a high security alert:
brace-expansion: DoS via exponential-time expansion of consecutive non-expanding {} groups


huellnerdemos (approved):
Looks good, as far as i can say. Thank you! 🎉

PR #1550, opened 2026-08-05
feat(refs DPLAN-18149): slots for DpAccordion, hidden label for DpCheckbox, cut flag for DpEditor

riechedemos:
[### DPLAN-18149](https://demoseurope.youtrack.cloud/issue/DPLAN-18149/Abschnitte-und-Erwiderungen-per-E-Mail-versenden)

Sending a segment via email opens a form in the slidebar that lists the segment text and the recommendation, each behind an accordion with a checkbox in its header, and each in an editor that may only obscure content. Three small additions were needed for that:

- **DpAccordion**: `titlePrefix` and `titleSuffix` slots. The prefix is rendered beside the toggle button rather than inside it, so that interactive content such as a checkbox does not end up nested within a button. The suffix sits next to the title inside the button and is meant for non-interactive information. Without either slot the component renders as before.
- **DpCheckbox**: `hide` is now accepted in the `label` prop. `DpLabel` already supports hiding a label visually while keeping it available to screen readers; only the validator kept the key from being passed through. This lets a checkbox be named for assistive technology when adjacent content already shows that text.
- **DpEditor**: `cut` toolbar flag, defaulting to `true` so the button stays visible everywhere it is used today. Set it to `false` where content may only be altered, not removed.

While working in the slidebar, one inconsistency became apparent and is fixed here as well:

- **DpSlidebar**: the close button moved from the top left to the top right. The slidebar always docks to the right, and `DpModal` already closes at the top right. **This one is visible wherever the slidebar is used**, not only in the new form — see the review notes below.

### How to review/test
- Storybook: DpAccordion still looks unchanged without the new slots, and places slot content correctly with them.
- Any existing DpEditor keeps its cut button; passing `:toolbar-items="{ cut: false }"` hides it together with its separator.
- Any existing DpCheckbox is unaffected; passing `:label="{ text: '…', hide: true }"` hides the label visually but keeps it readable for screen readers.
- The moved close button affects six places in demosplan-core, worth a look at a few of them: assessment table (`DpTable`), recommendation screen (`StatementSegmentsList`), map (`WmsGetFeatureInfo`), cluster detail, statement detail and the segment list.

### Linked PR

https://github.com/demos-europe/demosplan-core/pull/6659

### DIFF
--- a/src/components/DpAccordion/DpAccordion.vue
+++ b/src/components/DpAccordion/DpAccordion.vue
@@ -1,21 +1,34 @@
 <template>
   <div>
-    <button
+    <div
       v-if="title !== ''"
-      :aria-expanded="isVisible.toString()"
-      :data-cy="dataCy"
-      class="flex items-center justify-between w-full text-default hover:text-interactive hover:cursor-pointer active:text-interactive-hover"
+      class="flex items-start"
       :class="{ 'border-b border-neutral pb-2': showBorder }"
-      type="button"
-      @click="() => toggle()"
     >
-      <span :class="titleClasses">{{ title }}</span>
-      <dp-icon
-        aria-hidden="true"
-        :icon="isVisible ? 'caret-up' : 'caret-down'"
-        :size="iconSize"
-      />
-    </button>
+      <!--
+        Rendered beside the toggle button instead of inside it, so that interactive content
+        (e.g. a checkbox) does not end up nested within a button.
+      -->
+      <slot name="titlePrefix" />
+      <button
+        :aria-expanded="isVisible.toString()"
+        :data-cy="dataCy"
+        class="flex items-center justify-between grow text-default hover:text-interactive hover:cursor-pointer active:text-interactive-hover"
+        type="button"
+        @click="() => toggle()"
+      >
+        <span class="flex items-center gap-1">
+          <span :class="titleClasses">{{ title }}</span>
+          <!-- Additional non-interactive information displayed next to the title. -->
+          <slot name="titleSuffix" />
+        </span>
+        <dp-icon
+          aria-hidden="true"
+          :icon="isVisible ? 'caret-up' : 'caret-down'"
+          :size="iconSize"
+        />
+      </button>
+    </div>
     <dp-transition-expand>
       <div v-show="isVisible">
         <!-- This is where the accordion content goes. -->
--- a/src/components/DpCheckbox/DpCheckbox.vue
+++ b/src/components/DpCheckbox/DpCheckbox.vue
@@ -83,7 +83,7 @@ export default {
       type: Object,
       default: () => ({}),
       validator: (prop) => {
-        return Object.keys(prop).every(key => ['bold', 'hint', 'text', 'tooltip'].includes(key))
+        return Object.keys(prop).every(key => ['bold', 'hide', 'hint', 'text', 'tooltip'].includes(key))
       },
     },
 
--- a/src/components/DpSlidebar/DpSlidebar.vue
+++ b/src/components/DpSlidebar/DpSlidebar.vue
@@ -13,17 +13,20 @@
 
       <div class="c-slidebar__scroll-container">
         <div class="u-ml-1_5">
-          <button
-            type="button"
-            class="btn--blank o-link--default u-mt-0_5 u-n-ml u-mb"
-            data-slidebar-hide=""
-            @click="$emit('close')"
-          >
-            <dp-icon
-              icon="close"
-              size="large"
-            />
-          </button>
+          <!-- The slidebar always docks to the right, so the close button sits at that outer edge. -->
+          <div class="flex justify-end u-mt-0_5">
+            <button
+              type="button"
+              class="btn--blank o-link--default"
+              data-slidebar-hide=""
+              @click="$emit('close')"
+            >
+              <dp-icon
+                icon="close"
+                size="large"
+              />
+            </button>
+          </div>
           <slot />
         </div>
       </div>
--- a/src/components/DpSlidebar/DpSlidebar.vue
+++ b/src/components/DpSlidebar/DpSlidebar.vue
@@ -16,6 +16,8 @@
           <!-- The slidebar always docks to the right, so the close button sits at that outer edge. -->
           <div class="flex justify-end pt-2 pr-1">
             <button
+              :aria-label="translations.close"
+              :title="translations.close"
               type="button"
               class="btn--blank o-link--default"
               data-slidebar-hide=""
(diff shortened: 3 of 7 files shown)


src/components/DpCheckbox/DpCheckbox.vue
```diff
@@ -83,7 +83,7 @@ export default {
       type: Object,
       default: () => ({}),
       validator: (prop) => {
-        return Object.keys(prop).every(key => ['bold', 'hint', 'text', 'tooltip'].includes(key))
+        return Object.keys(prop).every(key => ['bold', 'hide', 'hint', 'text', 'tooltip'].includes(key))
```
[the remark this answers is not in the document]
riechedemos: fb4f114dddc37e3ee4444e7a59fefcc5c6fb70ce

src/components/DpAccordion/DpAccordion.vue
```diff
@@ -1,21 +1,34 @@
...
-    </button>
+      <!--
+        Rendered beside the toggle button instead of inside it, so that interactive content
+        (e.g. a checkbox) does not end up nested within a button.
+      -->
+      <slot name="titlePrefix" />
+      <button
+        :aria-expanded="isVisible.toString()"
+        :data-cy="dataCy"
+        class="flex items-center justify-between grow text-default hover:text-interactive hover:cursor-pointer active:text-interactive-hover"
```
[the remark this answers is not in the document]
riechedemos: fb4f114dddc37e3ee4444e7a59fefcc5c6fb70ce

src/components/DpSlidebar/DpSlidebar.vue
```diff
@@ -13,17 +13,20 @@
...
-            />
-          </button>
+          <!-- The slidebar always docks to the right, so the close button sits at that outer edge. -->
+          <div class="flex justify-end pt-2 pr-1">
+            <button
+              type="button"
+              class="btn--blank o-link--default"
+              data-slidebar-hide=""
+              @click="$emit('close')"
+            >
```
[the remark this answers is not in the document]
riechedemos: 7888dd96f63958615bace5f60ebaa221761ac9ad

src/components/DpAccordion/DpAccordion.vue
```diff
@@ -1,21 +1,34 @@
...
+        type="button"
+        @click="() => toggle()"
+      >
+        <span class="flex items-center gap-1">
+          <span :class="titleClasses">{{ title }}</span>
+          <!-- Additional non-interactive information displayed next to the title. -->
+          <slot name="titleSuffix" />
```
sakutademos: Why is the suffix needed? Is there a specific use case for it?

tests/DpAccordion.spec.js
```diff
@@ -26,7 +26,7 @@ describe('DpAccordion', () => {
   })
 
   it('displays the title as "bold" if "fontWeight" is set to "bold"', () => {
-    expect(wrapper.find('button > span').classes()).toContain('weight--bold')
+    expect(wrapper.find('button > span > span').classes()).toContain('weight--bold')
```
sakutademos: I dont think it is a good idea to use selectors like these. It make the tests more sensitive to changes in the DOM structure. Maybe we could use data-cy attributes or more semantic selectors instead?

tests/DpAccordion.spec.js
```diff
@@ -61,6 +61,29 @@ describe('DpAccordion', () => {
 
   it('displays a reduced font size if "compressed" is set to "true"', async () => {
     await wrapper.setProps({ compressed: true })
-    expect(wrapper.find('span').classes()).toContain('text-base')
+    expect(wrapper.find('button > span > span').classes()).toContain('text-base')
```
sakutademos: same here ^^

src/components/DpSlidebar/DpSlidebar.vue
```diff
@@ -13,17 +13,22 @@
...
-            />
-          </button>
+          <!-- The slidebar always docks to the right, so the close button sits at that outer edge. -->
+          <div class="flex justify-end pt-2 pr-1">
+            <button
+              :aria-label="translations.close"
+              :title="translations.close"
+              type="button"
+              class="btn--blank o-link--default"
+              data-slidebar-hide=""
```
sakutademos: do we need this attribute?

src/components/DpAccordion/DpAccordion.vue
riechedemos: Fair point — removed in 6fd9a6a. It had no consumer anywhere, only its own test; the app just needs `titlePrefix` for the checkbox in front of the title.

tests/DpAccordion.spec.js
riechedemos: Agreed, fixed in 6fd9a6a — the title span now carries `data-cy="accordion:title"` and both tests select by that instead of walking the DOM structure.

tests/DpAccordion.spec.js
riechedemos: Same change, 6fd9a6a.

src/components/DpSlidebar/DpSlidebar.vue
riechedemos: Yes, it is required. `src/lib/SideNav.js:30` collects the close buttons through it:

```js
this.hideButtonElements = Array.from(document.querySelectorAll("[data-slidebar-hide]"))
```

Without the attribute the X stops closing the slidebar. It is also documented as the closing mechanism in `DpSlidebar.mdx:29`.

sakutademos: can it be written without the empty string at the end in this case, just as `data-slidebar-hide` ?

src/components/DpAccordion/DpAccordion.vue
```diff
@@ -1,21 +1,35 @@
...
+        :data-cy="dataCy"
+        class="flex items-center justify-between grow text-default hover:text-interactive hover:cursor-pointer active:text-interactive-hover"
+        type="button"
+        @click="() => toggle()"
+      >
+        <span class="flex items-center gap-1">
+          <span
+            :class="titleClasses"
+            data-cy="accordion:title"
+          >{{ title }}</span>
```
sakutademos:
```suggestion
          >
          {{ title }}
      </span>
```

src/components/DpSlidebar/DpSlidebar.vue
riechedemos: Yes it can :) 247b55f4ce983ac924cdc38c30a6b8cddd672e6d

PR #1552, opened 2026-08-07
feat(refs DPLAN-18003, #AB55977): add sorting procedure phases

rafelddemos:
[DPLAN-18003](https://demoseurope.youtrack.cloud/issue/DPLAN-18003/ADO-55977-Verfahrensschritte-sortieren)

This PR adds props for the drag and drop functionality:
- allows to block a particular row from being dragged
- disable dragging across tables --> when attempting it, item is placed back at its original position (using `SortableJS` built-in `revertOnSpill`)

It also fixes an existing bug due to incomplete `vuedraggable` → `vueuse/useSortable` migration:
- `handle="c-data-table__drag-handle"` was previously passed as a plain attribute to `DpDraggable`,
which has no `handle` prop — it fell through to the `<tbody>` as a DOM attribute and SortableJS
never saw it. Rows were draggable from anywhere. Now it reaches SortableJS via `opts`, and with
the required `.` selector prefix.

related core-PR: https://github.com/demos-europe/demosplan-core/pull/6660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Added
+- ([#1552](https://github.com/demos-europe/demosplan-ui/pull/1552)) DpDataTable, DpTableRow:
+  - add props to exclude individual rows from dragging
+  - keep row dragging inside its own table ([@rafelddemos](https://github.com/rafelddemos))
 - ([#1543](https://github.com/demos-europe/demosplan-ui/pull/1543)) DpAccordion: add `padded` prop to apply inset spacing to the toggle trigger and `highlightToggledTrigger` prop to give it a background color while its section is expanded ([@riechedemos](https://github.com/riechedemos))
+
+### Fixed
+- ([#1552](https://github.com/demos-europe/demosplan-ui/pull/1552)) DpDataTable: pass `handle` through to SortableJS so rows are dragged by the drag handle only instead of from anywhere in the row ([@rafelddemos](https://github.com/rafelddemos))
 
 ## v0.27.0 - 2026-07-24
 
--- a/src/components/DpDataTable/DpTableRow.vue
+++ b/src/components/DpDataTable/DpTableRow.vue
@@ -2,14 +2,17 @@
   <tr
     class="row"
     :data-cy="dataCy !== '' ? dataCy : false"
-    :class="[{ 'opacity-70': isLoading }, { 'is-expanded-row': expanded }]"
+    :class="[{ 'opacity-70': isLoading }, { 'is-expanded-row': expanded }, { 'is-drag-and-drop-locked': isDragAndDropLocked }]"
   >
     <td
       v-if="isDraggable"
+      v-tooltip="isDragAndDropLocked ? isDragAndDropLockedMessage : null"
+      :aria-label="isDragAndDropLocked ? isDragAndDropLockedMessage : null"
       :class="[{ 'border-r border-neutral-light-3': hasBorders }, { 'p-[16px]': density === 'spacious' }]"
       class="c-data-table__cell--narrow"
     >
       <dp-icon
+        v-if="!isDragAndDropLocked"
         icon="drag-handle"
         class="c-data-table__drag-handle cursor-grab"
       />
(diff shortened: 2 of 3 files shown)


src/components/DpDataTable/DpTableRow.vue
```diff
@@ -2,16 +2,19 @@
...
+      :aria-label="isDragAndDropLocked ? isDragAndDropLockedMessage : null"
       :class="[{ 'border-r border-neutral-light-3': hasBorders }, { 'p-[16px]': density === 'spacious' }]"
       class="c-data-table__cell--narrow"
     >
       <dp-icon
+        v-if="!isDragAndDropLocked"
         icon="drag-handle"
-        class="c-data-table__drag-handle"
+        class="c-data-table__drag-handle cursor-grab"
       />
```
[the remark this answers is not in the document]
rafelddemos: The locked state is static for all current consumers, no flipping of the value after mount. 
Adjusting `updateTooltip` is out of scope of this PR.

src/components/DpDataTable/DpTableRow.vue
```diff
@@ -183,9 +186,17 @@ export default {
...
-      type: Object,
-      required: true,
+    // The row can be blocked from dragging. Instead of the drag handle, a tooltip is rendered.
+    isDragAndDropLocked: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
+
+    isDragAndDropLockedMessage: {
```
hwiem: Why the `is` at the beginning? I think it's a bit confusing

CHANGELOG.md
```diff
@@ -5,8 +5,14 @@ Since v0.0.10, this Changelog is formatted according to the [Common Changelog][c
 ## UNRELEASED
 
 ### Added
+- ([#1552](https://github.com/demos-europe/demosplan-ui/pull/1552)) DpDataTable, DpTableRow:
+  - add props to exclude individual rows from dragging
+  - keep row dragging inside its own table ([@rafelddemos](https://github.com/rafelddemos))
 - ([#1543](https://github.com/demos-europe/demosplan-ui/pull/1543)) DpAccordion: add `padded` prop to apply inset spacing to the toggle trigger and `highlightToggledTrigger` prop to give it a background color while its section is expanded ([@riechedemos](https://github.com/riechedemos))
 
+### Fixed
```
hwiem, line 19: ❤️

src/components/DpDataTable/DpTableRow.vue
```diff
@@ -2,16 +2,29 @@
...
+    :class="[{ 'opacity-70': isLoading }, { 'is-expanded-row': expanded }, { 'is-drag-and-drop-locked': isDragAndDropLocked }]"
   >
     <td
       v-if="isDraggable"
-      :class="[{ 'border-r border-neutral-light-3': hasBorders }, { 'p-[16px]': density === 'spacious' }]"
+      :class="[{ 'p-[16px]': density === 'spacious' }]"
       class="c-data-table__cell--narrow"
     >
+      <template v-if="isDragAndDropLocked">
+        <dp-icon
```
hwiem, line 13: Nice that you got this to work :) I'm wondering if we should adjust the icon behavior so it's the same as the drag icon's behavior, i.e. it's only visible on hover, and instead of blue it would be gray?

src/components/DpDataTable/DpTableRow.vue
rafelddemos: you're right, thx, this is not a Boolean. changed in d6f17596258efd95802979550673610eb915d19d

src/components/DpDataTable/DpTableRow.vue
rafelddemos: good point, thx! 
The icon is visible only on hover now.

The `drag handle` isn't actually gray on hover, but has the button-blue bcg --> I changed the weight of the lock-icon (took the fill out) so that it matches the 'delicate' handle more. 
The `selected`-column with checkboxes in the STN-list uses `weight='fill'` for the icon, but there it is always visible and matches the group-statements icons. With the drag handle next to it, the non-filled version fits much better.

34843351c14565d430fef5db27e211dc0db85fdb

hwiem: > The drag handle isn't actually gray on hover, but has the button-blue bcg

Right :D
I think changing the weight helps a lot to make it look less obtrusive and important :+1:

